PR #40, opened 2023-03-14
Move DDEV-Local -> DDEV and remove DDEV-Live

rfay:
Mostly

* DDEV-Local -> DDEV
* Remove DDEV-Live references

But I did get carried away in some other edits.


rfay:
So amazingly nice to be able to make significant edits this way. Amazing work @mattstein.

PR #43, opened 2023-04-17
Add homepage architecture callout

mattstein:
## The Issue

Mayank pointed out [in Discord](https://discord.com/channels/664580571770388500/1096577998712487936/1096577998712487936) that one of DDEV’s strengths is its platform support, which includes macOS M1 and M2 architecture. We don’t point that out anywhere on the homepage despite it being important!

This adds a prominent architecture callout to the homepage, using “Intel” and “Apple” next to “macOS” to specifically identify those platforms. (This seemed like it’d be more maintainable than listing out M1, M2, and continuing to tack on M* labels as new chips are released.)

Screen Shot 2023-04-17 at 09 25 17 AM@2x

[Visit the preview build](https://feature-homepage-architectur.ddev-com-front-end.pages.dev) and see what you think! My goal was to keep it from being too loud, but reinforce the supported architectures in a way that’s as legible and accessible as anything else on the page.

### DIFF
--- a/src/components/Badge.astro
+++ b/src/components/Badge.astro
@@ -1,0 +1,3 @@
+<div class="inline-block border rounded text-xs mx-[2px] px-1 py-0 relative bottom-px">
+  <slot />
+</div>
--- a/src/components/Hero.astro
+++ b/src/components/Hero.astro
@@ -2,11 +2,13 @@
 /**
  * The large, leading copy and image used on the homepage.
  */
-import CtaButton from "../components/CtaButton.astro"
+import CtaButton from "@components/CtaButton.astro"
+import Icon from "@components/Icon.astro"
 import AnimatedTerminal from "./AnimatedTerminal.astro"
+import { CloudIcon } from "@heroicons/react/24/solid/index.js"
 ---
 
-<div class="mt-12 py-32 relative overflow-x-hidden">
+<div class="mt-12 pt-24 pb-32 relative overflow-x-hidden">
   <div class="z-10 max-w-4xl mx-auto relative">
     <div class="max-w-2xl px-6 lg:px-0">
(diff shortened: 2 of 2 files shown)


rfay (changes requested):
Thanks so much for this! 

I think it's probably not clear enough showing both apple architectures, 

image


People *might* relate to "Apple Silicon" and "Intel", but I know that long word is what you were trying to avoid

I think they'd all relate to "M1/M2/Intel". How about that? 

Also note that DDEV runs great on Linux/arm64. Should be be showing that as well?

That of course means it works great on WSL2 on an arm64 machine, but that's a very small slice of the world.

mattstein:
@rfay It wouldn’t be a big deal to list “M1”, “M2”, and “Intel” right now, as I wrote above I assumed that would eventually get out of hand as new chips are released.

> Also note that DDEV runs great on Linux/arm64. Should be be showing that as well?
> 
> That of course means it works great on WSL2 on an arm64 machine, but that's a very small slice of the world.

If we need to call out chip architecture for each platform, I suppose we might as well. I specifically added badges to clarify Apple Silicon vs. Intel since that was @mayankguptadotcom’s primary concern, and it seemed like they were necessary for “Cloud” since that’s an otherwise nebulous term.

mattstein:
> I do think that just the base "Apple" after "macOS" is too ambiguous for most people.

Fair enough! I’m honestly not sure, but “M1”, “M2”, and “Intel” seem clear enough and some future version could describe “Apple Silicon” with a `title` attribute that spells out M1, M2, M3 support. Don’t need to make that today’s problem.

Updated!

rfay (approved):
Looks great to me, thanks!

mattstein:
Cool, I’ll merge this one too! 🚀

PR #63, opened 2023-07-11
Fix typos.

mattstein:
## The Issue

This fixes an incorrect date and capitalizes “Zoom”. Merge away if you approve!


rfay (approved):
Nobody could say you don't look at the finest details. 

I thought I was being really careful with those links. Thanks!

PR #72, opened 2023-07-26
Experimental: Zoho CRM Form for contributor page

rfay:
## The Issue

Attempt to replace google form with Zoho CRM Form.


rfay:
Contributor training blog page remains a mess here; still has the failed form in it. 
* Fix the form
* Consider moving the whole thing to a top-level training page.

rfay:
Form is fixed. https://20230726-new-form.ddev-com-front-end.pages.dev/blog/contributor-training/

rfay:
I think I'll wait on moving the contributor-training out of the blog until we start thinking about making it a regular thing (and capturing the content in it).

junk.html
mattstein, line 1: Is this file intentionally part of the PR?

src/components/Footer.astro
```diff
@@ -16,9 +16,10 @@ import Icon from "./Icon.astro"
 
 const navigation = {
   support: [
-    { name: "Documentation", href: DOCUMENTATION_URL },
-    { name: "GitHub Issues", href: GITHUB_ISSUES_URL },
+    { name: "Docs", href: DOCUMENTATION_URL },
+    { name: "Issues", href: GITHUB_ISSUES_URL },
```
mattstein, line 20:
```suggestion
    { name: "Docs", href: DOCUMENTATION_URL },
    { name: "Issues", href: GITHUB_ISSUES_URL },
```

Is there a reason to shorten “Documentation”, and make “GitHub Issues” _less_ specific when that’s still what it links to?

src/components/Footer.astro
```diff
@@ -16,9 +16,10 @@ import Icon from "./Icon.astro"
...
     { name: "Discord", href: DISCORD_URL },
+    { name: "Email", mailto: "support (at) ddev.com" }
```
mattstein, line 22: `mailto` isn’t a convention here, so you’ll see that this link doesn’t do anything. Even if it was, I’d be annoyed to have it open an email client and populate the `to:` field with an invalid address—wouldn’t you?

Rather than add links here, it might be more effective to create a catch-all Support page and link to that instead—like we do with the “Support DDEV” page as a clear call to action.

More links are more decision points and ultimately more noise, so it’s worth thinking through funnels and reinforcing them.

src/components/Footer.astro
```diff
@@ -45,14 +46,25 @@ const navigation = {
       me: true
     },
   ],
+  organization: [
```
mattstein, line 49: Is “organization” a meaningful new grouping? And is the newsletter organization-specific as this hierarchy might suggest?

You might consider moving both of these into product—I’d sign up for the DDEV newsletter expecting to hear about the product, for example, and the organization supports it I think. (That content could easily live on the about page.)

src/components/Footer.astro
```diff
@@ -45,14 +46,25 @@ const navigation = {
...
+    {
+      name: "Foundation",
+      href: "foundation",
```
mattstein, line 52:
```suggestion
      href: "/foundation",
```

src/components/Footer.astro
```diff
@@ -45,14 +46,25 @@ const navigation = {
...
+    },
+    {
+      name: "Newsletter",
+      href: "newsletter",
```
mattstein, line 56:
```suggestion
      href: "/newsletter",
```

src/pages/foundation.astro
```diff
@@ -0,0 +1,51 @@
...
+<Layout
+    title={title}
+>
+  <main class="max-w-4xl mx-auto mb-24">
+    <Heading title={`DDEV Foundation`} />
+    <div class="lg:flex">
+      <div class="px-6 lg:px-0 lg:w-2/3">
+        <PostBody>
+          <p>
+            The DDEV Foundation, a Colorado (USA) nonprofit corporation, is the developer of [DDEV](https://ddev.readthedocs.io). It was formerly named the Localdev Foundation. It serves as the fiscal entity for the project and owns the bank account, etc.
```
mattstein, line 18: Two issues with this link:

1. This is plain markup, so a Markdown link won’t work. (You can see that on the page.)
2. I would argue that the canonical internet home for DDEV is either ddev.com or github.com/ddev/ddev, so linking to the docs seems strange and unexpected here.

(I’d just as well point the ddev/ddev repository to ddev.com instead of the docs, since ddev.com should theoretically describe DDEV clearly and link to different interests like project support, technical support, and documentation. Same thing I mentioned earlier about intentional channeling and funneling.)

src/pages/foundation.astro
```diff
@@ -0,0 +1,51 @@
...
+          </p>
+          <p>
+            The DDEV Foundation will be applying for US 501(c)(3) non-profit status in 2023.
+          </p>
+
+          <h2>Where does the money go?</h2>
+
+          <p>
```
mattstein, line 26: Might be a good place to link to the about page since this is a little redundant.

src/pages/about.astro
```diff
@@ -61,8 +61,8 @@ const releaseData = await getReleases()
             an <a
               href="https://github.com/ddev/ddev/discussions/categories/ddev-advisory-group"
               >Advisory Group</a
-            >, and funded via the <a href="https://localdev.foundation"
-              >Localdev Foundation</a
+            >, and funded via the <a href="foundation"
```
mattstein, line 64:
```suggestion
            >, and funded via the <a href="/foundation"
```

src/pages/newsletter.astro
```diff
@@ -0,0 +1,31 @@
...
+    description={`Sign up for our newsletter.`}
+>
+  <main class="max-w-4xl mx-auto mb-24">
+    <Heading title={`Sign up for our newsletter`} />
+
+    This is a low-volume newsletter, usually just once a month. We'd love to be able to keep in touch with you.
+
+    <!-- NEW FORM GOES BELOW HERE -->
+
+    <iframe width='610px' height='210px' src='https://crm.zoho.com/crm/WebFormServeServlet?rid=d761f49b3421baca337046aa6e88b357c00ca36d6d951c2b6cd512fb5452c9edgide888d842d2182432c1bf117aecd6280f1bf0b73e665551779217573985104582'></iframe>
```
mattstein, line 27: This thing has an unsavory look to it on the page with the vertical scroll bar and left padding, but I don’t know how much control you have over that.

src/pages/newsletter.astro
```diff
@@ -0,0 +1,31 @@
...
+---
+
+<Layout
+    title={title}
+    description={`Sign up for our newsletter.`}
+>
+  <main class="max-w-4xl mx-auto mb-24">
+    <Heading title={`Sign up for our newsletter`} />
+
+    This is a low-volume newsletter, usually just once a month. We'd love to be able to keep in touch with you.
```
mattstein, line 23: I think it’s nice to quantify how frequently it goes out. Might be good to say what it’s about so I know whether I want to subscribe. (Is it enough to keep up with Discord announcements, or follow blog posts, for example, or is there something new or unique about the newsletter?)

src/pages/foundation.astro
```diff
@@ -0,0 +1,51 @@
...
+          <p>Contact support (at) ddev.com</p>
+
+          <h2>Snail Mail and Phone:</h2>
+          <pre>
+          DDEV Foundation
+          848 Montclair Dr
+          Palisade, CO 81526
+          USA
+          +1 970-462-7450
+          </pre>
```
mattstein, line 41: Why’s this in a code block?

If you definitely want that, you can use Astro’s [`set:html`](https://docs.astro.build/en/reference/directives-reference/#sethtml) to add precisely the text with newlines you want and drop the current indentation that’s a little strange.

But it might as well be a paragraph with `<br>` elements, or you could get fancier and do readers a favor using the [`<address>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address) element.

mattstein (changes requested):

mattstein:
> Updated footer; added "org" section; "Social" somehow never got implemented, so added that:

It’s implemented!

Screen Shot 2023-08-07 at 09 04 52 AM@2x

rfay:
> It’s implemented!

I guess a little obscure for me. Both places may be better. Existing is pretty, but I'd never see it. Thanks for chasing it down. I sure expected it to be with its friends.

junk.html
rfay: I think you know the answer :) Usually I'm better at avoiding things like that.

src/components/Footer.astro
rfay: The longer word "Documentation" messed up the footer columns when there are 4 of them. I imagine you know something better to do than shorten the word, happy to know what to do!

src/components/Footer.astro
rfay: I was just avoiding having a long list on one place. 

I don't think people read the about page long enough to discover anything, but would think it needs a link there anyway.

src/pages/foundation.astro
rfay: My basic goal was to move a full statement about the foundation to one place, the content is from localdev.foundation, and the idea is to replace that website. Sometimes a variety of things want a link to the *organization* page for documentation reasons, and they want the answers to their questions on the one page, not as links.

src/pages/newsletter.astro
rfay: I didn't like that much either. So on the next round (blog/contributor-training) I made it a js thingie, see if you like that one better.

src/pages/newsletter.astro
rfay: More important is to figure out how to allow unsubscribe, which currently is "reply to randy".

src/pages/foundation.astro
rfay: This came over from ddev.foundation.  Your suggestion is welcome.

mattstein:
> I guess a little obscure for me. Both places may be better. Existing is pretty, but I'd never see it. Thanks for chasing it down. I sure expected it to be with its friends.

Using social icons in the footer is a common (and I would argue expected) pattern, and duplicating that effort leaves more for the visitor to parse that’s particularly unfriendly to anyone with screen reader that would step over the same thing twice. That’s the last I’ll say about it though, so take that however you will!

src/components/Footer.astro
rfay: Please let me know what you'd do here to just give away the email address without giving it away to the world.

src/components/Footer.astro
mattstein: I intentionally used the longer word in the footer since the shortened version was already in the header. I’d always have the page layout support the content and not the other way around, so adjust the columns instead. (If there needs to be another column, per other comments.)

src/components/Footer.astro
rfay: Removed organization and put those under product.

src/components/Footer.astro
mattstein: My experience has been that I’ll collect spam regardless of anything I could possibly do, and it [seems like](https://stackoverflow.com/questions/11563283/why-write-at-and-dot-in-email-rather-than-and) the [jury’s out](https://www.quora.com/Is-writing-out-the-at-in-email-addresses-effective-to-prevent-spam?share=1) on “mailbox (at) domain” formatting being effective.

I’d recommend any kind of obfuscation that isn’t a penalty to the visitor. (Because `mailto:support (at) ddev.com` would put the burden on me to fix it in my email client in order to send mail to `support@ddev.com`.)

[This article](https://spencermortensen.com/articles/email-obfuscation/) suggests that either URL encoding or CSS would be viable approaches.

src/components/Footer.astro
rfay: Tried changing it back to "Documentation" but don't like it, happy to have a suggestion that will make it work. 

image

src/components/Footer.astro
rfay: Changed to use URL encoding, thanks so much!

src/pages/foundation.astro
rfay: Fixed, sorry, markdown is in my blood now and copy/paste. 

> (I’d just as well point the ddev/ddev repository to ddev.com instead of the docs, since ddev.com should theoretically describe DDEV clearly and link to different interests like project support, technical support, and documentation. Same thing I mentioned earlier about intentional channeling and funneling.)

Didn't understand your statement here, but if you were saying "We should use ddev.com as the standard reference when talking about the project" I agree.

src/pages/newsletter.astro
rfay: Changed to the scriptlet approach

src/pages/foundation.astro
rfay: Added link to About page though.

rfay:
Thanks for all the super-helpful review and suggestions. I think everything's mostly the way I'd like it, but we haven't resolved the footer. I don't like what happens with "Documentation" there. I do want to normal footers (in addition to the icons). Although the icons are WAY OUT OF DATE thanks to our "friend" crazy Elon.  I can't ever find twitter links in my stuff now.

src/pages/foundation.astro
mattstein: We’re agreeing then—I’m referencing other places where a whole bunch of links could be a simple link to ddev.com for a more clear path to other things:
Screen Shot 2023-08-07 at 10 10 24 AM@2x
Screen Shot 2023-08-07 at 10 11 00 AM@2x

rfay: absolutely.  Those are done.

mattstein:
I don’t agree that duplicating information is the way to go, but if it’s down to column widths being the problem you can adjust the Tailwind classes:

```html
<div class="grid grid-cols-auto gap-8 xl:col-span-4">
  <div class="md:grid md:grid-cols-4 md:gap-8">
```

That’ll give them breathing room without being awkwardly wide.

rfay:
Thanks, I updated that and it looks good. Looks like those are still built for 4 columns though even though we're back to 3 columns?

image

mattstein:
Yep! That’s what I meant by breathing room without being awkwardly wide. Make it three columns and the amount of negative space between columns makes the whole thing feel noisy and scattered. With that change, they still read as a visual unit.

rfay:
I was asking about `xl:col-span-4` and `md:grid-cols-4` which seem to imply 4 columns. If it's all OK it's all OK!

src/components/Footer.astro
rfay: I think a contact form shouldn't be hard and should be in our future. Will leave it with the obfuscation right now.

rfay:
Thanks for all your help on this. I guess Tailwind is yet another skill I'm missing. Didn't recognize those attributes at all :)

And of course, thanks for caring for it and seeking consistency, clarity, and proper communication.

PR #73, opened 2023-07-27
Add Contributor Training page and supporting content

mattstein:
## The Issue

The attempt to add a signup form with inline CSS, JavaScript, and markup within Markdown content isn’t going very well in #72. Using a dated blog post as a persistent source of information also stretches the purpose of dated posts that should probably be reserved for topical adventures and announcements.

The timeanddate.com links are also helpful but cumbersome to manage.

> 👀 Build preview: https://feature-training-section.ddev-com-front-end.pages.dev/training/

Since contributor training would probably be an ongoing thing, this creates a dedicated page and content type for it.

This means each session is represented by a little `.md` file in `src/content/trainings/`:

```md
---
title: Testing Infrastructure
sessionStart: "2023-09-19T14:00:00Z"
sessionMinutes: 60
recordingUrl:
---
```

This information is all that’s necessary to generate the page you’ll see in the build preview. `sessionStart` is in UTC, the mother of all timezones, and `sessionMinutes` is included because currently there’s no indication up front how long these sessions are supposed to be. `recordingUrl` requires a valid URL and can be added later.

Each field gets validated, and I rewrote titles to maintain Title Case and try to be succinct—with one exception that demonstrates how content is presented when you add it to the Markdown file. (Which should probably be used sparingly.)

#### timeanddate.com links are generated automatically and include the full session title. 

Screen Shot 2023-07-26 at 07 36 30 PM@2x

#### A `title` attribute is included on the date column to reiterate the timing and length of the session.

Maybe overkill, but it’s hard to misinterpret!

Screen Shot 2023-07-26 at 07 35 37 PM@2x

#### A persistent callout on the “Support DDEV” page linking to this new one without being obnoxious.

Screen Shot 2023-07-26 at 07 37 01 PM@2x

#### Clicking “Register” next to an upcoming event prefills the “Session” input and scrolls to the form.

Screen Shot 2023-07-26 at 07 38 00 PM@2x

### The Form

I don’t know how this part is supposed to work, so I made up a quick + ideal flow:

1. Added a field for designating whatever session someone’s interested in, because otherwise how do you know which they should be invited to?
2. Consolidated “First Name” and “Last Name” into “Name” to speed things up. (Does it matter if you have someone’s first name only if you mainly need their email address?)
3. Removed “Company”, assuming that’s not pertinent to the contributor training signup.

It doesn’t do anything yet, and I didn’t bother copying all the cruft in from the Zoho CRM snippet.

Its narrower breakpoints are also lame at the moment, so yikes on mobile.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,6 +29,7 @@
         "glob": "^10.2.2",
         "image-size": "^1.0.2",
         "linkedom": "^0.14.21",
+        "luxon": "^3.3.0",
         "marked": "^4.2.5",
         "mdast-util-to-string": "^3.1.0",
         "nanoid": "^4.0.1",
@@ -45,6 +46,8 @@
         "tailwindcss": "^3.2.4"
       },
       "devDependencies": {
+        "@tailwindcss/forms": "^0.5.4",
+        "@types/luxon": "^3.3.1",
         "@types/marked": "^4.0.8",
         "dotenv": "^16.0.3",
         "textlint": "^13.0.5",
@@ -1852,6 +1855,18 @@
         "escalade": "^3.1.1"
       }
     },
+    "node_modules/@tailwindcss/forms": {
+      "version": "0.5.4",
+      "resolved": "https://registry.npmjs.org/@tailwindcss/forms/-/forms-0.5.4.tgz",
+      "integrity": "sha512-YAm12D3R7/9Mh4jFbYSMnsd6jG++8KxogWgqs7hbdo/86aWjjlIEvL7+QYdVELmAI0InXTpZqFIg5e7aDVWI2Q==",
+      "dev": true,
+      "dependencies": {
+        "mini-svg-data-uri": "^1.2.3"
+      },
+      "peerDependencies": {
+        "tailwindcss": ">=3.0.0 || >= 3.0.0-alpha.1"
+      }
+    },
     "node_modules/@tailwindcss/typography": {
       "version": "0.5.9",
       "resolved": "https://registry.npmjs.org/@tailwindcss/typography/-/typography-0.5.9.tgz",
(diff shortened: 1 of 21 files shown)


rfay:
You're awesome!

mattstein:
> 1. Would this support adding additional content to the trainings? For example, they should probably have outlines, perhaps full blog posts, perhaps links to new content in the docs. Perhaps embedded video.

They technically support more content now, but I only built it with at most a sentence that lives on this page—so it’d get awkward and distracting cramming content into the table. Each session could just as well have its own page similar to a blog post, so we’d just need a URL structure that works.

Which means we’d also need to be clear about what each post represents: do you want a page _per training_ that you update each quarter/year with the latest training info, or a page _per session_?

> 2. The form... you have it here as a standalone, will that do the job for other things? One immediate need is a newsletter signup form. The raw-crazy code pasted from zoho is in my PR. Would love to have you copy from there and see if it can work here.

The form’s just markup here and it doesn’t do anything, because I don’t understand how Zoho is set up or how you intend for registration and scheduling to work. (Or how/why Zoho is involved, really.)

We can add a newsletter signup form, but that seems like it’d be entirely separate from this effort. We could surely use similar form markup, but I think that’s a moot point.

mattstein:
On second thought, I’m probably just forcing more questions, reverse-engineering goals, and adding complexity with this when a [Typeform](https://www.typeform.com/) link on the current blog page would be a more straightforward way to solve the problem.

rfay:
The key thing that I don't know how to do right now is to have a plain form that is allowed to do its deed without interference. I just want a plain HTML page based on what the CRM provides, so that contacts get added automatically into the CRM.  I don't think Typeform is the answer to that is it?

rfay:
Ah, Typeform does claim integration with Zoho CRM, https://www.typeform.com/connect/zoho-crm/

rfay:
Unfortunately Typeform costs as much or more than Zoho CRM, https://www.typeform.com/pricing/

mattstein:
> The key thing that I don't know how to do right now is to have a plain form that is allowed to do its deed without interference. I just want a plain HTML page based on what the CRM provides, so that contacts get added automatically into the CRM. I don't think Typeform is the answer to that is it?

You can add plain _markup_ within a blog post’s Markdown file and it should be rendered just fine. It’s when you layer on included inline JavaScript and CSS that things can start to go sideways. The Google form was easier because it was an iframe where all the innards were self-contained. So ignoring editorial intent, it’d be more ideal to create a `.astro` page like I did here (overbuilding the answer) because you have more freedom to tinker with the stuff you’re embedding. It acts more like a plain `.html` page would.

I suggested a Typeform link because it seems way more straightforward than fighting a form into place where the form may not need to exist. It could just as well be a standalone page or a standalone form—I haven’t worked with Zoho CRM specifically but I faintly recall others having form builders that could host the form pages for sharing.

rfay:
Astro is pretty great, isn't it!

mattstein:
I sure fancy it!

PR #75, opened 2023-08-09
Minor time-has-passed updates to tutorials

rfay:
## The Issue

Some minor things getting out of date.

Reviewing this as a result of reading the [Nix docs](https://nix.dev/tutorials/#tutorials), which have tutorials to start with, like you have here. I'd sure like to figure out how to incorporate the amazing work you've done here directly into the docs. Or docs directly into ddev.com (better). 

It's so important for people to have an easy startup experience, and right now the docs are scattered, as hard as they try. ddev.com is much better.


mattstein (approved):
Looks good to me!

I’d also love it if the docs could be pulled into ddev.com and easier to shape to suit their purpose like this page starts to exemplify. I don’t know how to pull that off though—my last job was focused mostly on this challenge, and it worked well but it was a full-time effort. Not just making it happen, but then maintaining the machinery.

PR #76, opened 2023-08-15
Implement foundation page as markdown, with MarkdownTemplate.astro

rfay:
## The Issue

Watching https://craftquest.io/courses/astro-quick-start-guide/50358 I was inspired to try to make top-level markdown work on ddev.com

Uses the basic approaches there along with @mattstein suggestions, especially the Tailwind "prose" class. It works OK at this point but is not ready for prime time.

* I couldn't get it to build if I added in the `<SEO>` component
* It's apparent to me that I don't know what I'm doing having MarkdownLayout.astro import Layout.astro

This can currently be run locally with `npm run dev` and http://localhost:3000/foundation is "not bad", but you'll want it to be mo betta. 

However, `npm run build` fails with

```
 error   Cannot read properties of null (reading 'getAttribute')
  File:
    /Users/rfay/workspace/ddev.com-front-end/src/lib/search-index.js:85:43
  Code:
    84 |             // Don’t include `noindex` pages in search results
    > 85 |             continue
         |                                           ^
      86 |           }
      88 |           const mainTag = postDocument.querySelector("main")
  Stacktrace:
TypeError: Cannot read properties of null (reading 'getAttribute')
```

Thanks for your help so far @mattstein - I'd be happy to book a session with you to look at this or also happy to receive async comments.


rfay:
May work on this again some day.

PR #83, opened 2023-08-15
Add link to today's ddev.com training in contributor-training

rfay:
## The Issue

Add link to today's ddev.com training.


rfay:
Now why didn't this one generate a preview I wonder. It's not on fork.

PR #86, opened 2023-08-21
docs: update codespaces install, fixes #85

### LINKED ISSUE #85: Update guide for codespaces, remove docker-in-docker
- After https://github.com/ddev/ddev/pull/5267 is merged, guide on homepage needs an update as well
- `docker-in-docker` is not needed anymore if the universal image is used

Guide: https://ddev.com/get-started/
image

Don't have the time resources in the next days to dive into astro (missed the contributor training unfortunately) and provide a PR, but I could provide it in the next weeks.

Just wanted to create the issue so that this does not get lost. Cheers!

rfay:
## The Issue

* #85 

Things always changing :)

Compare to https://ddev.readthedocs.io/en/latest/users/install/ddev-installation/#github-codespaces

image

### DIFF
--- a/src/components/quickstart/Cloud.astro
+++ b/src/components/quickstart/Cloud.astro
@@ -48,9 +48,9 @@ const { latestVersion } = Astro.props
           </li>
           <li>
             Edit the project’s <code>.devcontainer/devcontainer.json</code> file
-            to include these two lines in the “features” section:
+            to include this line in the “features” section:
             <Code
-              code={`"features": {\n    "ghcr.io/devcontainers/features/docker-in-docker:2": {},\n    "ghcr.io/ddev/ddev/install-ddev:latest": {}\n},`}
+              code={`"features": {\n    "ghcr.io/ddev/ddev/install-ddev:latest": {}\n},`}
               theme="nord"
               lang="json"
             />


rfay:
Please take a look @mandrasch

mandrasch:
> I failed again and did this on my fork.

Hey! Sure thing.

What fork did you use exactly? https://github.com/rfay/ddev with main branch? The main branch of your fork does not have a `.devcontainer/devcontainer.json` yet as far as I could see. Therefore project configuration does not exist.

image

When I use https://github.com/ddev/ddev, it offers me "Go" (because I forgot to change remove the name in line https://github.com/ddev/ddev/blob/master/.devcontainer/devcontainer.json#L4C2-L4C2). But this is correct and uses universal image.

image

But odd thing is port `8080` is not used:

image

image

The project is instead running on https://laughing-guacamole-9p595g9gph7q6-32771.app.github.dev/ `32771` 🤔 

Regarding the "get started" guide in general:

Should we we provide a simple example repository? 

I quite liked the idea of the `d9-simple` sample project inside ddev repo, but maybe something more easy like index.php with "hello world!" would also be sufficient to get the core idea of DDEV + php in the cloud? Something like `ddev/codespaces-demo` e.g. which uses the `devcontainer.json` of https://ddev.readthedocs.io/en/latest/users/install/ddev-installation/#github-codespaces?

Or we provide the full working `.devcontainer/devcontainer.json` which people can copy into their project
(https://github.com/mandrasch/ddev-craftcms-vite/tree/main/.devcontainer)? 

Both ways would be helpful I guess ...

mandrasch:
Regarding the "get started" guide in general:

Should we we provide a simple example repository? 

I quite liked the idea of the `d9-simple` sample project inside ddev repo, but maybe something more easy like index.php with "hello world!" would also be sufficient to get the core idea of DDEV + php in the cloud? Something like `ddev/codespaces-demo` e.g. which uses the `devcontainer.json` of https://ddev.readthedocs.io/en/latest/users/install/ddev-installation/#github-codespaces?

Or we provide the full working `.devcontainer/devcontainer.json` which people can copy into their project
(https://github.com/mandrasch/ddev-craftcms-vite/tree/main/.devcontainer)? 

Both ways would be helpful I guess ...

mandrasch:
Or we just provide the link to https://github.com/ddev/d9simple, just realized by now that this is already a separate repository. :-)

mandrasch:
> Or we just provide the link to https://github.com/ddev/d9simple, just realized by now that this is already a separate repository. :-)

Updated the d9simple repo here: https://github.com/mandrasch/d9simple/tree/feat-remove-docker-in-docker, can submit this as PR. 

Selected `mandrasch/d9simple` and `feat-remove-docker-in-docker` branch:
image

Launching this via "create new codespace" worked fine. 

Just need to add the db import scripts as well from https://github.com/ddev/ddev/blob/master/.devcontainer/setup_test_project.sh, if this should be used as official example:
```
    DDEV_ARTIFACTS=${DDEV_REPO}-artifacts
    git clone ${DDEV_ARTIFACTS} "/tmp/${DDEV_ARTIFACTS##*/}" || true
    ...
    ddev import-db --file=/tmp/${DDEV_ARTIFACTS##*/}/db.sql.gz
    ddev import-files --source=/tmp/${DDEV_ARTIFACTS##*/}/files.tgz
    ...
```

rfay:
> I failed again and did this on my fork.

Sorry, I just meant that to get the proper preview here in ddev.com-front-end I can't use a fork when doing a PR because the automated build doesn't happen then. I gave an image from my local though.

rfay:
Now... I didn't understand if you approved of this and it accomplished your goal or not.

mandrasch:
Hi, sorry for confusion, was a bit late yesterday in austria. 😉  😴 

Review: 

Needs more work. 

I would propose we take a step back, the current guide on ddev.com can be confusing for new users since some crucial information is missing. 

Especially this part is missing:

image

A good first solution for me would be to copy the updated tutorial from docs to ddev.com as well: https://ddev.readthedocs.io/en/latest/users/install/ddev-installation/#github-codespaces 

With that we would not have to verify two different guides.

rfay:
Let's finish the in-play work on 
* https://github.com/ddev/ddev/pull/5290

then and we'll come back to this here. 

My basic thinking is that we should make it much simpler over there. It's already so much harder than gitpod, and we need to fine tune what we're asking them to do.

PR #93, opened 2023-08-25
Add blog post for August 22, 2023 contributor training

kristin-wiseman:
## The Issue
I created a blog post for the contributor training on Tuesday (8/22/23)

## Automated Testing Overview
I ran textlint... It didn't like the phrase "not just answers", and I sort of struggled to come up with a suitable replacement for the word "just". 

@mattstein if you have any suggestions to improve my writing in this post, please let me know because I appreciate your input. @rfay - All of the original content from your outline is there, but I reorganized it and tried to elaborate on some points by including some of your spoken content too. Please let me know if I missed something important?

### DIFF
--- a/src/content/blog/how-to-give-and-get-community-support.md
+++ b/src/content/blog/how-to-give-and-get-community-support.md
@@ -1,0 +1,53 @@
+---
+title: "How to Give and Get Community Support"
+pubDate: 2023-08-24
+summary: Randy Fay's advice for giving and receiving support in the DDEV community
+author: Randy Fay
+categories:
+  - Community
+---
+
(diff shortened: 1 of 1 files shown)


rfay:
:) The blanket prohibition on "just" is there "just for me" because I just use that word all the time and it just drives @mattstein crazy..

rfay:
Thanks for this, trying to ignore some things for the weekend, but it will get in, thanks!

rfay (approved):
Awesome, thanks so much! 

One nit that I'd ask for future PRs (almost anywhere) is to try to make your branch name have more meaningful information. For example, you used the name `blog6` here, but something like `20230827_kristin-wiseman-support-blog` would be so much more helpful. That's the pattern that I recommend in https://ddev.readthedocs.io/en/latest/developers/building-contributing/#pull-requests-and-pr-preparation - remember that maintainers are often going to check out your branch, and then they will probably have that branch around locally and they'll never remember what `blog6` is. And you may not either. 

Thanks so much for translating this informal live session into useful exhortations for the future!

PR #107, opened 2023-09-29
Upgrade Astro to v3

mattstein:
## The Issue

Astro 3 [is out](https://astro.build/blog/astro-3/) and we should stay current for general maintenance and fun.

This upgrades all the npm dependencies, fixes some spacing issues I noticed, and uses the native assets library for the homepage GitHub images. It shouldn’t affect much else, including any v2 updates @bmartinez287 is working on.

Astro compresses HTML by default now, which is a good thing, but it means anything that relied on fragile whitespace in templates could get smooshed. I saw this in the footer and fixed it and didn’t see any other cases, but worth mentioning just in case I missed any.

n/a

## Automated Testing Overview

n/a

## Related Issue Link(s)

none

## Release/Deployment Notes

n/a


bmartinez287:
looks good

mattstein:
Alrighty, merging it!

rfay:
Thanks!

PR #108, opened 2023-10-04
Add mobile first spacing to hero header.

### LINKED ISSUE #103: Hero h1 margin and padding on mobile widths.
Before

Screen Shot 2023-09-24 at 7 27 00 PM

After

Screen Shot 2023-09-24 at 7 23 41 PM

There seems to be too much margin and padding on mobile devices.
At around 640px width there does not seem to be a need for the top margin and top padding to be so large.

Should I proceed with a PR and change the top margin and padding until it exceeds 640 which is tailwindcss sm breakpoint

bmartinez287:
## The Issue

Closes #103 

It adds smaller margins and paddings on mobile devices.

### DIFF
--- a/src/components/Hero.astro
+++ b/src/components/Hero.astro
@@ -9,7 +9,7 @@ import AnimatedTerminal from "./AnimatedTerminal.astro"
 import { CloudIcon } from "@heroicons/react/24/solid/index.js"
 ---
 
-<div class="mt-12 pt-24 pb-32 relative overflow-x-hidden">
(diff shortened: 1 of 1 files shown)


mattstein:
I’m not sure if we’re waiting on me, but it looks great so merge away if @rfay is happy with it. Thanks for fixing this, @bmartinez287!

bmartinez287:
Since we have approval above, I'm merging it in.

PR #110, opened 2023-10-04
Add a blog edit link

bmartinez287:
## The Issue
The docs have a quick edit button that directs one straight to the docs but the blog does not.
This PR tries to address that.

Screen Shot 2023-10-04 at 7 34 59 PM

I used the example the Astro docs had and placed it on what seemed to be the most convenient place.

### DIFF
--- a/src/components/Heading.astro
+++ b/src/components/Heading.astro
@@ -17,9 +17,11 @@ export interface Props {
   authorAvatar?: string
   // Human-formatted read time (used for blog posts)
   readTime?: string
+  // Human-formatted read time (used for blog posts)
+  editUrl?: string
 }
 
-const { title, subtitle, date, author, authorUrl, authorAvatar, readTime } =
+const { title, subtitle, date, author, authorUrl, authorAvatar, readTime, editUrl } =
   Astro.props
 ---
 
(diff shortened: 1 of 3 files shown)


bmartinez287:
The wording for edit this page comes from the Astro docs too but I'm good with just calling it edit if needed or something else.

rfay:
So needed, thanks!

mattstein:
I like this as an addition and the wording’s great! I’d suggest giving it a visual treatment that makes it slightly less prominent and more clearly indicates that it’s a link instead of another piece of post meta information. (Maybe float it off to the right, use the text font instead of mono, and make sure it has a hover effect using opacity or underline?) I’d nudge the pencil icon down a bit so it’s optically aligned with the text, but that’s just starting to get picky.

bmartinez287:
That sounds great. Thanks for the feedback the more the better, even if is a little adjustment. 
It helps spec out the features and have a more consistent look to them. I will work on those tweaks. The link one is a good one, I thought about only doing the pencil icon since we do that with the docs. 
That is simpler potentially but it might be easier to miss.

mattstein:
> The link one is a good one, I thought about only doing the pencil icon since we do that with the docs.
That is simpler potentially but it might be easier to miss.

I wasn’t suggesting you get rid of the pencil, just make sure the visual treatment of the whole thing distinguishes it as an action to take rather than a bit of blog post information like the author and date.

rfay:
This is going to be a great step, and next time I know you'll put it in a branch in ddev/ddev.com.

But right now I wanted to test it, and `npm run dev` doesn't work for me any more

```
$ npm run dev

> @ddev/ddev.com-astro@1.0.0 dev
> astro dev

05:20:52 PM [astro] Unable to load /Users/rfay/workspace/ddev.com/astro.config.mjs
```

bmartinez287:
I only seen that lately if I forget to do 'npm install'. But I will check on it.

rfay:
I'll bet I forgot, thanks.

bmartinez287:
Screen Shot 2023-10-15 at 1 25 50 PM

A picture of the new version with it.

bmartinez287:
I think, I covered all requests but I will do one more pass to check that everything works as expected.

src/components/Heading.astro
```diff
@@ -64,10 +66,34 @@ const { title, subtitle, date, author, authorUrl, authorAvatar, readTime } =
...
             {" "}
-            <span aria-hidden="true"> ·</span> {readTime}
+            <span aria-hidden="true"> · </span> {readTime}
           </span>
         )
       }
     </p>
+    {
+      editUrl && (
+        <a class="flex dark:text-white pl-1 sm:pl-4 items-center sm:ml-auto font-serif hover:underline" href={editUrl} target="_blank">
```
mattstein, line 76: Was there a reason for `font-serif` here as opposed to `font-sans`? This would be the only place on the site where a serif font shows up, so it seems a little out of place to me.

mattstein (changes requested):
Left a question about the serif font but otherwise it looks great!

src/components/Heading.astro
bmartinez287: Not in particular, I thought that was the other main one. Must have overseen that.

bmartinez287:
It's been switched.

mattstein (commented):
Cool, thanks for doing that!

mattstein (approved):

rfay:
Yay!

PR #111, opened 2023-10-09
Add ddev config file

bmartinez287:
## The Issue

This repo does not use a ddev file that requires one to install npm. I tried making my own a while back and ended up down a rabbit hole. It potentially adds a barrier to contrib.
Because ddev is already on my machine it would be nice to take advantage of it to among other things switch npm versions and have easy access to a database.

This PR adds the ddev config file. However, I ran into a few issues.

```
Network ddev_default created
Failed to get project(s): invalid app type: astro

````
If I set the project type to astro it will complain since it's not on the list and that makes sense I suppose but It made me think whether there should be a custom type or a generic type.

Additionally, I was trying to run ddev npm run dev and map the local HTTPs server port 4321 but when I tried it with  `web_extra_exposed_ports` it did not work. It gave me a bad route error.

I might be missing some history on why some of these items are the way they are. I will look on Discord and see if I can find them.

### CODE REVIEW

**Comment by bmartinez287** [comment]
After looking at some plugins I wonder if Astro dev settings should be part of a plugin such as vite and others.
https://github.com/tyler36/ddev-viteserve
The vite one is a good example. That might still require a custom project type or some sort of tag for a project base.

File: .ddev/config.yaml [hunk 1]
```diff
@@ -0,0 +1,256 @@
+name: ddev.com
```
rfay, line 1: It's important not to use a real on-internet name in these situations. It causes enormous grief and confusion.

However, I usually recommend leaving out the `name` when providing a pre-populated config.yaml so it can just use the project name of the enclosing directory.

```suggestion
```

.ddev/config.yaml
```diff
@@ -0,0 +1,256 @@
...
+type: laravel
```
rfay, line 2: I think `php` will be much better than `laravel`. `php` is essentially "generic"

```suggestion
type: php
```

.ddev/config.yaml
```diff
@@ -0,0 +1,256 @@
...
+docroot: dist
+webserver_type: nginx-fpm
+router_http_port: "80"
+router_https_port: "443"
```
rfay, line 6: Neither of these are in config.yaml by default any more. 

```suggestion
```

.ddev/config.yaml
```diff
@@ -0,0 +1,256 @@
...
+additional_hostnames: []
+additional_fqdns: []
+mutagen_enabled: false
```
rfay, line 9: This is not a current config option. 

```suggestion
performance_mode: none
```

.ddev/config.yaml
```diff
@@ -0,0 +1,256 @@
...
+# can have their intended affect. 'override_config' affects only behavior of the
+# config.*.yaml file it exists in.
+
+# Many ddev commands can be extended to run tasks before or after the
+# ddev command is executed, for example "post-start", "post-import-db",
+# "pre-composer", "post-composer"
+# See https://ddev.readthedocs.io/en/stable/users/extend/custom-commands/ for more
+# information on the commands that can be extended and the tasks you can define
+# for them. Example:
+#hooks:
```
rfay, line 256: IMO the comments aren't useful here.

```suggestion
```

rfay (changes requested):
Thanks - The next thing to add is to update the README explaining how to use it. 

You could also get just slightly fancier and change the nginx config to use astro as its upstream instead of php-fpm, and then everything would be on ports 80 and 443.

rfay (approved):
This seems fine. If you wanted to automatlcally `npm run dev` you could do it with a `web_extra_daemons` stanza.

bmartinez287:
I'm looking into the nginx config to switch from php-fpm but I'm not sure if I found the right settings there.

Additionally, Astro runs Vite underhood so I figure maybe the Vite ddev add-on or running vite as a web_extra_exposed_ports would do it but so far I haven't been able to get HMR Hot module reloading working with Astro Vite implementation. I saw a few support posts on discord about some getting it to work but they are using a standard vite install and I think Astro does a wrapper which might complicate it a bit.

The dev.config file works well as it is if one just runs `npm run build ` but it be nice to solve it for `npm run dev` as well

bmartinez287:
I get an astro.conf in /etc/supervisor/conf.d/astro.config after I do a ddev.ssh with the following settings.
```
[program:astro]
group=webextradaemons
command=bash -c "npm run dev || sleep 2"
directory=/var/www/html
autostart=false
autorestart=true
startretries=15
stdout_logfile=/var/tmp/logpipe
stdout_logfile_maxbytes=0
redirect_stderr=true
```
Which seems fine. Looking at this issue https://github.com/ddev/ddev/issues/4872 but port 9998 or 9999 does not let me reach it it gives me a bad gateway.

I imagine this should be reachable http://ddev.com.ddev.site:9998/ and I thought that would give me the same output as `npm run dev` on a local environment. I must be missing a step since I got a bad gateway error.

bmartinez287:
time="2023-10-12T23:55:54Z" level=debug msg="'502 Bad Gateway' caused by: dial tcp 192.168.112.4:4321: connect: connection refused"
192.168.112.1 - - [12/Oct/2023:23:55:54 +0000] "GET /favicon.ico HTTP/2.0" 502 11 "-" "-" 51 "ddev-astrosite-web-4321-https@file" "http://ddev-ddev-astrosite-web:4321" 1ms

bmartinez287:
Got it working, the issue was a missing -- that Atro Vite required to allow all hosts.
I added instructions to the README and updated the files.

.ddev/config.yaml
```diff
@@ -0,0 +1,35 @@
+name: astrosite
```
rfay, line 1: Recommend just leaving this out if possible. It will then take the name of the directory it's in.

```suggestion
```

.ddev/config.yaml
```diff
@@ -0,0 +1,35 @@
...
+web_extra_exposed_ports:
+- name: astro-dev
+  container_port: 4321
+  http_port: 4322
+  https_port: 4321
+
+# The extra -- is a Vite item
+# https://github.com/vitejs/vite/discussions/3396
+web_extra_daemons:
+  - name: astro-dev-deamon
```
rfay, line 22: Might as well spell it as expected. 
```suggestion
  - name: astro-dev-daemon
```

.ddev/config.yaml
```diff
@@ -0,0 +1,35 @@
...
+  post-start:
+    - exec: echo '================================================================================='
+    - exec: echo '                                  NOTICE'
+    - exec: echo '================================================================================='
+    - exec: echo 'The Astro dev container is ready.'
+    - exec: echo
+    - exec: echo "By default it will be available at port 4321 - URL:4321"
+    - exec: echo "Use ddev describe to confirm if this doesn't work."
+    - exec: echo 'Check the status of startup by running "ddev logs --follow --time"'
+    - exec: echo '================================================================================='
```
rfay, line 35: * It will be a lot easier to make these one single exec, use newlines. 
* It might make more sense to use `exec-host` for this as well, since there's no need that I know of to execute inside the container.

README.md
```diff
@@ -53,20 +53,33 @@ All commands are run from the root of the project, from a terminal:
...
 
+
+
 ### Local Development Setup
 
+1. Run `cp .env.example .env` to create a `.env` file for environment variables. (Don’t check this in!)
+2. Create a [classic GitHub access token](https://github.com/settings/tokens) with these scopes: `repo`, `read:org`, `read:user`, and `read:project`.
+3. Paste the GitHub token after `.env`’s `GITHUB_TOKEN=`.
+
+#### DDev setup
```
rfay, line 64:
```suggestion
#### DDEV setup
```

README.md
```diff
@@ -53,20 +53,33 @@ All commands are run from the root of the project, from a terminal:
...
+
+Ddev already has all the dependencies included.
```
rfay, line 66:
```suggestion
DDEV already has all the dependencies included.
```

README.md
```diff
@@ -53,20 +53,33 @@ All commands are run from the root of the project, from a terminal:
...
 
 1. Run `nvm use` to make sure you’re running an appropriate Node.js version.
 2. Run `npm install` to set up the project’s dependencies.
-3. Run `cp .env.example .env` to create a `.env` file for environment variables. (Don’t check this in!)
-4. Create a [classic GitHub access token](https://github.com/settings/tokens) with these scopes: `repo`, `read:org`, `read:user`, and `read:project`.
-5. Paste the GitHub token after `.env`’s `GITHUB_TOKEN=`.
-6. Run `npm run dev` to start Astro’s dev server.
-7. Visit the URL displayed in your terminal. (Probably `http://127.0.0.1:3000/`.) The site will automatically refresh as you work on it, displaying errors in the relevant console (terminal or browser).
+3. Run `npm run dev` to start Astro’s dev server.
+4. Visit the URL displayed in your terminal. (Probably `http://127.0.0.1:4321/`.) The site will automatically refresh as you work on it, displaying errors in the relevant console (terminal or browser).
```
rfay, line 80: This should be `https://<projectname>.ddev.site:4321`. Maybe 127.0.0.1 works anyway? I doubt it, because then ddev-router doesn't know where to route it to.

README.md
```diff
@@ -53,20 +53,33 @@ All commands are run from the root of the project, from a terminal:
...
+To generate a static copy of the site, run `ddev npm run build`. The contents of the `dist/` folder are exactly what get [deployed to Cloudflare Pages](#build--deployment). You can preview locally by running `ddev npm run preview` or using a tool like [`serve`](https://www.npmjs.com/package/serve).
+
+#### General setup
+
 Check out the project in your favorite Node.js environment, ideally running [`nvm`](https://github.com/nvm-sh/nvm). We’ll install dependencies, add a GitHub API key, and run a local dev server with a hot-reloading browser URL.
 
 1. Run `nvm use` to make sure you’re running an appropriate Node.js version.
 2. Run `npm install` to set up the project’s dependencies.
-3. Run `cp .env.example .env` to create a `.env` file for environment variables. (Don’t check this in!)
-4. Create a [classic GitHub access token](https://github.com/settings/tokens) with these scopes: `repo`, `read:org`, `read:user`, and `read:project`.
```
rfay, line 63: Isn't the GITHUB_TOKEN still needed to download the assets?

rfay (changes requested):
Congrats on all the good work. Variety of suggestions, I haven't manually tested yet.

README.md
bmartinez287: it is but since we can run it with both ddev and just npm on the host instructions I extracted that and added it as the first few steps when setting up the local environment. So is higher up before ddev setup.

The current order is 
Local development instructions.
- Github token

Then based on that one chooses between the DDev setup or the general Setup.

README.md
bmartinez287: This is for the local setup without ddev. Since some might choose that if they have npm already installed.

.ddev/config.yaml
bmartinez287: Sounds good

README.md
rfay: OK, so I was reading too fast and without adequate context? Please make sure the two sections are well-separated and clear what's what. (They probably are. It's just the problem of reading diffs!)

README.md
```diff
@@ -53,20 +53,33 @@ All commands are run from the root of the project, from a terminal:
...
+1. Run `cp .env.example .env` to create a `.env` file for environment variables. (Don’t check this in!)
+2. Create a [classic GitHub access token](https://github.com/settings/tokens) with these scopes: `repo`, `read:org`, `read:user`, and `read:project`.
+3. Paste the GitHub token after `.env`’s `GITHUB_TOKEN=`.
+
+#### DDev setup
+
+DDev already has all the dependencies included.
+
+1. Run `ddev start && ddev npm install` set up the project’s dependencies.
+3. `npm run build` can be found at the base URL and `npm run dev` is found at URL:4321. Dev has Vite HMR(hot module reloading) among other features. The site will automatically refresh as you work on it, displaying errors in the relevant console (terminal or browser).
```
rfay, line 69:
```suggestion
2. `ddev npm run build` can be found at the base URL and `ddev npm run dev` is found at URL:4321. Dev has Vite HMR(hot module reloading) among other features. The site will automatically refresh as you work on it, displaying errors in the relevant console (terminal or browser).
```

rfay (changes requested):
`ddev npm run build` gets

```
▶ src/pages/index.astro
node:internal/event_target:1033
  process.nextTick(() => { throw err; });
```


Since you're using web_extra_exposed_ports you shouldn't be telling people to hit http://127.0.0.1, should be named URL, for example `https://ddev.com.ddev.site:4321` (which works)

But `https://ddev.com.ddev.site:4321/blog` gets me a 404. 

The post-start hook shows `By default it will be available at port 4321 - URL:4321`, but there's no reason it can't show the full URL using `${DDEV_PRIMARY_URL}:4321`

README.md
bmartinez287: ${DDEV_PRIMARY_URL}:4321 -  I was looking for that variable for a while haha, I will work on that.

http://127.0.0.1/ is for the local setup without DDev in case they just want to run npm but maybe I can make that clearer I will double check.

I did not see that build error I will check on that one as well.

Thanks for the feedback.

bmartinez287:
I updated my OS to Sonoma, switched to Colima, and have been running into an odd issue where it gives me an operation not permitted.
Screenshot 2023-10-16 at 3 36 39 PM

Screenshot 2023-10-16 at 3 41 15 PM

I'm probably hitting an OS bug now :expressionless:

That with `ddev start && ddev npm cache clean --force  && ddev npm install` the fact one has to cache clean as sometimes it gets stuck in between. 
More updates to follow.

bmartinez287:
Screenshot 2023-10-16 at 4 05 51 PM

ddev npm run build works well for the most part.

rfay:
Not sure what you were hitting there, especially if you had mutagen enabled. Can't tell much about what you were running against with Colima without knowing what Colima config you were using and whether you had mutagen enabled. However, note that Lima just had an update...

bmartinez287:
https://ddev.readthedocs.io/en/stable/users/install/docker-installation/#macos
colima start --cpu 4 --memory 6 --disk 100 --vm-type=qemu --mount-type=sshfs --dns=1.1.1.1
I'm doing brew docker version 24.0.6 and  brew colima version 0.5.6.

Yeah, I might have gone too close to the bleeding edge now.  O well, it will work eventually hahaha.

rfay:
Not sure how you can get permissions errors with that setup *if using mutagen*. 

New limactl version (which makes a big difference) is
```
limactl version 0.18.0
```

bmartinez287:
Ups, I was not using mutagen. Thanks for adding that note.

bmartinez287:
It's working!!! and  I think is ready.

To run it in ddev make sure to run `ddev start  && ddev npm install` so the daemon won't crash the web container.

If it fails then run `ddev npm cache clean --force && ddev npm install && ddev restart `

README.md
```diff
@@ -65,7 +65,7 @@ All commands are run from the root of the project, from a terminal:
 
 DDev already has all the dependencies included.
 
-1. Run `ddev start && ddev npm install` set up the project’s dependencies.
+1. Run `ddev start && ddev npm cache clean --force  && ddev npm install ` set up the project’s dependencies.
```
mattstein: Safe to get rid of the extra space unless that was important:

```suggestion
1. Run `ddev start && ddev npm cache clean --force  && ddev npm install` set up the project’s dependencies.
```

README.md
```diff
@@ -65,7 +65,7 @@ All commands are run from the root of the project, from a terminal:
 
 DDev already has all the dependencies included.
```
mattstein, line 66: Not sure where “DDev” came from, but it [should always be referred to](https://ddev.readthedocs.io/en/stable/developers/writing-style-guide/#writing-style) as DDEV the product or `ddev` the command.

```suggestion
DDEV already has all the dependencies included.
```

README.md
```diff
@@ -65,7 +65,7 @@ All commands are run from the root of the project, from a terminal:
...
 
-1. Run `ddev start && ddev npm install` set up the project’s dependencies.
+1. Run `ddev start && ddev npm cache clean --force  && ddev npm install ` set up the project’s dependencies.
 3. `ddev npm run build` can be found at the base URL and `ddev npm run dev` is found at https://<projectname>.ddev.site:4321. Dev has Vite HMR(hot module reloading) among other features. The site will automatically refresh as you work on it, displaying errors in the relevant console (terminal or browser).
```
mattstein, line 69: 3 → 2, fix spacing, reword slightly for clarity:

```suggestion
2. `ddev npm run build` can be found at the base URL and `ddev npm run dev` is found at https://<projectname>.ddev.site:4321. The dev server has Vite HMR (hot module reloading) among other features. The site will automatically refresh as you work on it, displaying errors in the relevant terminal or browser console.
```

What do you mean by `ddev npm run build` can be found at the base URL?

I know that running the `build` command generates a static version of the site into the `dist/` directory, which you can run the `preview` command to serve, but the wording of this seems sort of ambiguous. Similarly with “`ddev npm run dev` is found at”—the command isn’t found at a browser address, but running it starts a server you can access from the browser at that address. I know you’re aware of this, but the wording leaves it unclear for someone figuring it out.

README.md
```diff
@@ -53,20 +53,39 @@ All commands are run from the root of the project, from a terminal:
...
+#### DDev setup
+
+DDev already has all the dependencies included.
+
+1. Run `ddev start && ddev npm install ` set up the project’s dependencies. If it fails then run `ddev npm cache clean --force && ddev npm install && ddev restart`. 
+3. `ddev npm run build` can be found at the base URL and `ddev npm run dev` is found at https://<projectname>.ddev.site:4321. Dev has Vite HMR(hot module reloading) among other features. The site will automatically refresh as you work on it, displaying errors in the relevant console (terminal or browser).
+
+To generate a static copy of the site, run `ddev npm run build`. The contents of the `dist/` folder are exactly what get [deployed to Cloudflare Pages](#build--deployment). You can preview locally by running `ddev npm run preview` or using a tool like [`serve`](https://www.npmjs.com/package/serve). 
+
+#### Without DDev setup
```
mattstein, line 73:
```suggestion
#### Setup Without DDEV
```

README.md
```diff
@@ -53,20 +53,39 @@ All commands are run from the root of the project, from a terminal:
...
 
 1. Run `nvm use` to make sure you’re running an appropriate Node.js version.
 2. Run `npm install` to set up the project’s dependencies.
-3. Run `cp .env.example .env` to create a `.env` file for environment variables. (Don’t check this in!)
-4. Create a [classic GitHub access token](https://github.com/settings/tokens) with these scopes: `repo`, `read:org`, `read:user`, and `read:project`.
-5. Paste the GitHub token after `.env`’s `GITHUB_TOKEN=`.
-6. Run `npm run dev` to start Astro’s dev server.
-7. Visit the URL displayed in your terminal. (Probably `http://127.0.0.1:3000/`.) The site will automatically refresh as you work on it, displaying errors in the relevant console (terminal or browser).
+3. Run `npm run dev` to start Astro’s dev server.  If it fails then run `npm cache clean --force && npm install && npm run dev`. 
+4. Visit the URL displayed in your terminal. (Probably `http://localhost:4321/`.) The site will automatically refresh as you work on it, displaying errors in the relevant console (terminal or browser).
```
mattstein, line 80:
```suggestion
4. Visit the URL displayed in your terminal. (Probably `http://localhost:4321/`.) The site will automatically refresh as you work on it, displaying errors in the relevant terminal or browser console.
```

README.md
```diff
@@ -53,20 +53,39 @@ All commands are run from the root of the project, from a terminal:
...
 
 To generate a static copy of the site, run `npm run build`. The contents of the `dist/` folder are exactly what get [deployed to Cloudflare Pages](#build--deployment). You can preview locally by running `npm run preview` or using a tool like [`serve`](https://www.npmjs.com/package/serve).
 
+
+#### Switching from without Ddev to with DDev local environment.
```
mattstein, line 85: Use correct “DDEV”s and shorten:

```suggestion
#### Switching from Without DDEV to With DDEV
```

README.md
```diff
@@ -53,20 +53,39 @@ All commands are run from the root of the project, from a terminal:
...
+
+Make sure to delete your node vendor folder everytime you switch and run npm install. The change in architecture can create odd issues otherwise.
```
mattstein, line 87: Be specific, use backticks for command:

```suggestion
Make sure to delete your `node_modules/` directory and run `npm install`. The change in architecture can create odd issues otherwise.
```

mattstein (changes requested):

bmartinez287:
I'm running into a few odd issues here among them.  
`ddev start && ddev npm install && ddev npm run build` might be too of an operation. It seems to get killed before it completes execution.

`ddev start && ddev npm install` works but for first-time setups, they would need the build folder to since that's where https://<projectname>.ddev.site takes you.

Additionally, web_extra_daemons sometimes runs before ddev npm install and that break the build as it won't have anything to build upon. `command: "sleep 3 && npm run dev -- --host"` I added the sleep 3 to help with it since sometimes it is a race condition that could be mitigated by it. Is there another way to make `ddev npm install` run before web_extra_daemons?

The post-start hook would not do it because that has the same race condition issue.

rfay:
If you're seeing `killed` check your Docker memory allocation.

bmartinez287:
This PR works but the daemon running before npm install has completed does concern me a bit. It fails if it can't find those files. I feel like I need a hook between pre-start and post-start so I can still do the npm install but it won't run the daemon until is over with it.

rfay:
I'll try to take a look at this before long.

rfay:
Sorry I lost track of this.

bmartinez287:
No worries, im sure you been pinned down after traveling fora while.

rfay:
Rebased

rfay:
I'm seeing this kind of behavior:
```
> @ddev/ddev.com-astro@1.0.0 dev
> astro dev --host

  🚀  astro  v3.4.4 started in 554ms

  ┃ Local    http://localhost:4321/
  ┃ Network  http://172.27.0.2:4321/
             http://172.26.0.4:4321/

04:48:20 PM [content] Watching src/content/ for changes
04:48:21 PM [content] Types generated
04:48:21 PM [astro] update /.astro/types.d.ts
2023-11-08 16:48:22,015 INFO waiting for child_exit_monitor, php-fpm, nginx, mailpit, astro-dev-daemon to die
2023-11-08 16:48:25,023 INFO waiting for child_exit_monitor, php-fpm, nginx, mailpit, astro-dev-daemon to die
2023-11-08 16:48:26,030 WARN killing 'astro-dev-daemon' (2493) with SIGKILL
2023-11-08 16:48:27,031 INFO stopped: astro-dev-daemon (terminated by SIGKILL)
2023-11-08 16:48:27,033 INFO stopped: mailpit (exit status 0)
2023-11-08 16:48:27,068 INFO stopped: nginx (exit status 0)
[08-Nov-2023 16:48:27] NOTICE: Terminating ...
[08-Nov-2023 16:48:27] NOTICE: exiting, bye-bye!
2023-11-08 16:48:27,079 INFO stopped: php-fpm (exit status 0)
2023-11-08 16:48:27,079 INFO stopped: child_exit_monitor (terminated by SIGTERM)
```

Anyway, it comes up but I think astro dies soon, and then the whole web container gets brought down. I've seen this lately running `npm run dev` on the host a few times as well, it starts up and then dies.

rfay:
Never mind. I didn't follow the instructions.

README.md
```diff
@@ -53,20 +53,39 @@ All commands are run from the root of the project, from a terminal:
...
-7. Visit the URL displayed in your terminal. (Probably `http://127.0.0.1:3000/`.) The site will automatically refresh as you work on it, displaying errors in the relevant console (terminal or browser).
+3. Run `npm run dev` to start Astro’s dev server.  If it fails then run `npm cache clean --force && npm install && npm run dev`. 
+4. Visit the URL displayed in your terminal. (Probably `http://localhost:4321/`.) The site will automatically refresh as you work on it, displaying errors in the relevant terminal or browser console.
 
 To generate a static copy of the site, run `npm run build`. The contents of the `dist/` folder are exactly what get [deployed to Cloudflare Pages](#build--deployment). You can preview locally by running `npm run preview` or using a tool like [`serve`](https://www.npmjs.com/package/serve).
 
+
+#### Switching from Without DDEV to with DDEV
+
+Make sure to delete your `node_modules/` directory and run `npm install`. The change in architecture can create odd issues otherwise.
```
rfay, line 87: (There is no change in architecture though, but probably a difference in node version, related problems.)
```suggestion
Make sure to delete your `node_modules/` directory and run `ddev npm install`. The change in architecture can create odd issues otherwise.
```

rfay (approved):
Working great for me, thanks! Minor suggestion in README.

README.md
bmartinez287: True, it could probably be clearer. I meant, npm downloads a different of the files locally as it does inside docker. I thought it was an arm64 vs non arm issue but that does not make much sense now I think about it. But it was something like that.

PR #121, opened 2023-10-16
Add copy button for code blocks on MD files - branch

bmartinez287:
The Issue
Currently, blog posts don't have a copy button snippets of code. The DDev docs do but not ddev.com

How This PR Solves The Issue
Astro uses the astro-expressive-code plugin which provides that functionality.
https://github.com/expressive-code/expressive-code/blob/main/packages/astro-expressive-code/README.md

The image below is taken from this blog
https://ddev.com/blog/amd64-on-apple-silicon-ddev/

Screen Shot 2023-10-15 at 10 55 44 PM


bmartinez287:
Here it is.

rfay (approved):
Working great, thanks!

bmartinez287:
So the faux window shown in the pic comes from this plugin.
https://www.npmjs.com/package/@expressive-code/plugin-frames
https://www.npmjs.com/package/@expressive-code/plugin-frames#terminal-window-frames
Code blocks will be rendered as a terminal window if their language identifier matches one of the supported terminal languages bash, shell script, shell, sh, or zsh:
That piece of code has it because it does a 
```
\```bash
#!/bin/bash
```
Same plugin that has the copy button

bmartinez287:
So after looking into this a little more the faux window seems more like a feature. I could go back through all 20 guides and change it so it looks appropriate if needed. But that's one feature that seems pretty bundled with it. Hard to turn off on its own. I would only show on specific cases in which one had bash, shell script, shell, sh, or zsh.

The shadow is gone. We could technically change the typeface but I'm not sure its worth it since they look pretty similar and Astro docs use the IBM one which is the one we are moving to.

Screenshot 2023-10-16 at 6 51 15 PM

I like how Astro as the file path on the top corner. Right now that feature is disabled because we haven't used it so it would display like this.

Screenshot 2023-10-16 at 6 52 42 PM

And yet I could see how that could be useful when trying to find where a file lives. But for backwards compatibility we can do that later or just do a pass on the blogs and get the file path and the faux window on or off.

bmartinez287:
Makes sense. We could leave it here until a better alternative arrives or maybe we can figure how to do full replacement with it.

rfay:
Where did we end up on this?

bmartinez287:
We could get it to work on markdown files, but the quick-start page works differently.  So we likely had to rework the quick-start page and make it into a markdown file instead of the current custom template.
Then, we might have to work around current and older blog posts to make them fully compatible with the new plugin syntax. They will work as they are, but for blogpost to be consistent we may have to modify some slightly.

rfay:
So there's no reasonable path for this right now I guess? Was a great initiative.

bmartinez287:
I'm looking at Astro new https://starlight.astro.build/getting-started/
starlight documentation theme and looking at what we can pull from there when it comes to the copy button and redoing some of the get started pages so they inherit the copy button and are not left behind.

rfay:
Would love to see this move forward.

rfay:
This is so very old that I'll close it, but would love to have this feature.

PR #136, opened 2023-11-08
Promote training sessions

tyler36:
This PR improves the visibility of the training sessions by:

- adding link in footer to blog post.
- adding a section on the contributing page.

### DIFF
--- a/src/components/Footer.astro
+++ b/src/components/Footer.astro
@@ -25,6 +25,7 @@ const navigation = {
   product: [
     { name: "About", href: "/about" },
     { name: "Contributing", href: "/support-ddev" },
+    { name: "Contributor Live Training", href: "/blog/contributor-training"},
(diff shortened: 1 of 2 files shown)


tyler36:
I think the past training links, future training links, and sign up form would be better suited on the contributing page instead of in a past blog post. The contributing page would then be updated moving forward. Thoughts?

rfay:
I agree that in the future we should probably have a better place for this in general.

mattstein:
I took a run at a dedicated page [over here](https://github.com/ddev/ddev.com/pull/73) before I realized I was probably overthinking it when the eternally-updated blog post seemed to be going okay.

rfay:
We're probably near the end of the run for this set of training sessions anyway, only one more is currently scheduled, so maybe next time we'll do it the right way.

PR #137, opened 2023-11-08
Highlight support patterns

tyler36:
This PR adds links to each support pattern on the contributing page.

- highlights at glance some methods
- enables links that can be shared


mattstein (approved):
Good idea! I’d use hyphens for word separators to be consistent, but that’s mostly nitpicky preference.

tyler36:
> I’d use hyphens for word separators to be consistent, but that’s mostly nitpicky preference.

Updated.

mattstein (approved):

rfay (approved):
Thanks!

PR #138, opened 2023-11-08
add table of content to readme

tyler36:
This PR adds a table of content to the readme.

The readme file is quite long, therefore a TOC is helpful for navigation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,21 @@
-# ddev.com Front End
+# ddev.com Front End <!-- omit in toc -->
 
 Source code for [ddev.com](https://ddev.com)’s static front end, built with [Astro](https://astro.build) to keep things organized, maintainable, and fast.
+
+- [Overview](#overview)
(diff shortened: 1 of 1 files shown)


mattstein:
@tyler36 I get that it’s nice to see a table of contents front and center, but are you sure it’s worth maintaining when every GitHub repo gets one for free?

Screen Shot 2023-11-07 at 06 13 29 PM@2x

tyler36:
... and there is my "Today I learned", thank you. 😄 

Valid point. I was viewing the project in VSCode, and not on the web.  

In VSCode, I use "[Markdown All in One](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one)" for all my project which can generate and auto-updates TOCs so maintainence isn't an issue, for me.

Closing.

mattstein:
Another platform lets you pop in a tag that auto-generates the TOC inline. Don’t want to name names, but it rhymes with LitGab. 🙂

Issue #141, opened 2023-11-09
Feed images


tyler36:
During tracking #140 , it was raised that ddev.com 's feed images were not appearing in [feedly ](https://feedly.com/)(rss reader).

![image](https://github.com/ddev/ddev.com/assets/7234392/950f4c64-a840-4f3d-9bf1-4dab67e209ba)

The top-level image also appears to be using a default image (same as "Matt Glaman" and "PHP" feeds for reference):

![image](https://github.com/ddev/ddev.com/assets/7234392/95ddd6de-b41d-4633-b1b1-dc68ea9eab88)
![image](https://github.com/ddev/ddev.com/assets/7234392/137b9e25-7b63-4203-9d70-3f93eea4f2b5)

We should investigate if this is a feedly issue or an issue with our feed.

mattstein:
The feed is _valid_, but we probably need to make updates to appease Feedly specifically.

The site’s favicon does not exist at the root (ddev.com/favicon.ico), for example, and I’d bet that’s exactly and only what Feedly is looking for. Similarly with those images, I assume Feedly is looking for a summary image we don’t have in place.

mattstein:
It turns out Feedly does have some [automatic and not-so-automatic things it looks for](https://blog.feedly.com/10-ways-to-optimize-your-feed-for-feedly/). Just merged a PR that should now include images and an icon.

tyler36:
Thank you @mattstein 

Feed icon is now displaying
![image](https://github.com/ddev/ddev.com/assets/7234392/932ea444-4ea9-4025-ae54-394e69ae6e2e)

I'm not seeing images for feed items. Not sure if this is a caching issue, or if they will start appearing with new items.
![image](https://github.com/ddev/ddev.com/assets/7234392/89f1d085-19bc-4ca6-9912-5885193e47b9)


I checked the feed on https://rss.app/ (online RSS reader) and images display there so the URL is right.

![image](https://github.com/ddev/ddev.com/assets/7234392/698694ae-afec-4c85-8dca-021ddfe2bf36)

mattstein:
I wasn’t seeing the icon yet @tyler36, so I guess that’s progress! 🤞

tyler36:
There was a new item in DDEV feed but the image did not show up in Feedly.

![image](https://github.com/ddev/ddev.com/assets/7234392/6faafa84-a0f4-41c9-a920-d989cf7f24f6)

[...]

Issue #140, opened 2023-11-08
Duplicate and invalid feed items


tyler36:
Today, my feed contained 3 new items from ddev.com

![image](https://github.com/ddev/ddev.com/assets/7234392/7395f71b-102d-4eff-84e0-d369d9432394)

There are several issues here:
- "Working with Vite in DDEV - an introduction" is repeated twice
- "Introducing New DDEV Maintainer Stas Zhuk" appeared in my feed 16 days ago, and is a duplicate.
- The 2 top feeds (5h) are invalid links (`https://ddev.comblog/`)
  - [Working with Vite in DDEV - an introduction](https://ddev.comblog/working-with-vite-in-ddev)
  - [Introducing New DDEV Maintainer Stas Zhuk](https://ddev.comblog/introducing-maintainer-stas)
  
  Update: I'm using https://feedly.com/ as my reader.

rfay:
Thanks for paying attention! Please take a look at https://ddev.com/blog/rss.xml (assuming that's what you're using) and see what you think. 

You might try `curl -s https://ddev.com/blog/rss.xml | xq` (assuming you have xq installed) to review it. 

A quick look didn't make me think there was anything wrong, but we appreciate your sharp eye.

tyler36:
Feedly is subscribed to `https://ddev.com/blog/rss.xml` (same as you suggested)
Eyeballing the feed, the first item looks strange, the link is incorrect.

```xml
<title>Working with Vite in DDEV - an introduction</title>
<link>https://ddev.comblog/working-with-vite-in-ddev</link>
<guid isPermaLink="true">https://ddev.comblog/working-with-vite-in-ddev</guid>
```

Perhaps feedly saw this as a "new" item and hence the duplicate? 
... scanning down the list, all the `link` seem to be missing the `/` between `.com` and `blog`.

rfay:
Want to see if you can sort it out? Sure looks wrong. Suggestions @mattstein ? (Matt has looked into this before). 

I don't know how the missing slash would break what items you're getting though, and don't know how anything can work with the slash missing.

BTW, I see the same thing you do in feedly:

<img width="1097" alt="image" src="https://github.com/ddev/ddev.com/assets/112444/2e782251-86f4-4f25-8397-a6a91475112a">

mattstein:
Whoops! Pretty sure I inadvertently broke the feed URLs [upgrading to Astro 3](https://github.com/ddev/ddev.com/pull/107), sorry!

[...]

PR #142, opened 2023-11-12
Spruce up RSS feed

mattstein:
## The Issue

#140 and #141 raised some issues with the RSS feed:

1. Item links and GUIDs have had invalid URLs since the upgrade to Astro 3. (Courtesy of my oversight.)
2. Post content may include relative links that carry through to the feed.
3. The Feedly experience includes duplicate posts because of item 1.
4. The Feedly experience is underwhelming without post images and an icon. 

I previously made a [quick change](https://github.com/ddev/ddev.com/commit/80476d0c976ebd7af48c88c4d059c56b2f38e522) to fix those invalid URLs (item 1).

This PR ensures that relative URLs in content are automatically converted to absolute ones (item 2).

The PR also ensures that feature images are included in the post content, which Feedly will use. It also adds a namespace just for Feedly in order to include an icon, logo, accent color, and generic cover image. All of this should amount to [nicer presentation in Feedly](https://blog.feedly.com/10-ways-to-optimize-your-feed-for-feedly/)—I tested to confirm what I could without the changes being live on ddev.com.

I don’t know if it’s possible to ask Feedly to ignore those bad URLs it previously got from the feed, but those are gone so the root cause is at least solved. Again, sorry for that being an issue!

You can add the RSS feed URL to Feedly to see it use feature images, but I think the reference logo and icons will need to exist on ddev.com to come through.


rfay:
Thanks for this!

Issue #146, opened 2023-11-20
Search URLs are all broken, search is unusable


rfay:
If you go to https://ddev.com/search/ for example, and search for "different" you'll get 
<img width="923" alt="image" src="https://github.com/ddev/ddev.com/assets/112444/3c8e8b18-5f90-4571-987d-4fabaaae84a8">

But if you click any of the links, they're all missing the hostname, so the top item links to `https://blog/whats-so-different-about-ddev-local/`, which of course doesn't work.

rfay:
@bmartinez287 do you have time to take an initial look at this?

bmartinez287:
I will take a look.

PR #147, opened 2023-11-21
Fixes DDEV search component.

bmartinez287:
## The Issue
The issue is described here
* https://github.com/ddev/ddev.com/issues/146

It adds the domain to the javascript search utility.

### DIFF
--- a/src/pages/search.astro
+++ b/src/pages/search.astro
@@ -43,9 +43,10 @@ const title = `Search`
 
     resultsContainer.innerHTML = results
       .map((result) => {
+        let resulturl = window.location.origin + result.id
         return `
         <a href="${
-          result.id
+          resulturl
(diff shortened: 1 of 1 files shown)


bmartinez287:
After looking at a few alternatives such as creating a constant, or saving the domain on the search.json file this seemed to be the quickest and potentially the most elegant fix. @rfay

rfay (approved):
Well, it sure works and it sure seems elegant! Thanks.

bmartinez287:
I will merge it in since it's already broken, but if you have a suggestion for a fix that's a bit better let us know. @mattstein. I think it originally worked because it made some assumptions that the recent major upgrade took away.

rfay:
OK, sounds like a plan. We can go back to it if Matt has a better idea.

bmartinez287:
No worries, thanks for keeping the project up to date.

PR #149, opened 2023-11-28
Docker provider blog

rfay:
## The Issue

We are supporting more Docker providers than ever, explain about the new ones, explain Docker, show usage.


src/content/blog/docker-providers.md
```diff
@@ -0,0 +1,58 @@
...
+
+## How can DDEV support all these providers?
+
+The idea of Docker is that everything happens inside a container, which you can think of as a little, separate computer. So the Docker images and containers used by DDEV are the same on every operating system, except that we build separate images for each [processor architecture](blog/arm64-apple-silicon-m1-ddev-local-what-does-it-all-mean/) (ARM64 - Apple Silicon, and AMD64 - Intel). So mostly all the work is done inside the container, and the containers/images behave the same. And DDEV’s main code is written in Golang, which is compiled natively into a single binary on every platform. Of course that sounds simple, but it’s not all that simple. Building all those images has quite a history, and of course Docker providers that ought to behave the same everywhere don’t actually, so there are workarounds in many places.
```
mattstein, line 40: Minor suggestions for flow, clarity, the concept of multiple containers, and reassurance that DDEV is the one managing workarounds so you don’t have to:

```suggestion
The idea of Docker is that everything happens inside a container, which you can think of as a little, separate computer. You can have as many of them as you want, and they can be created and destroyed easily.

DDEV’s Docker images and containers are the same on every operating system, except that we build separate images for each [processor architecture](blog/arm64-apple-silicon-m1-ddev-local-what-does-it-all-mean/) (ARM64 = Apple Silicon, and AMD64 = Intel). So mostly all the work is done inside the container, and the containers/images behave the same. And DDEV’s main code is written in Golang, which is compiled natively into a single binary on every platform. Of course that sounds simple, but it’s not all that simple. Building all those images has quite a history, and of course Docker providers that ought to behave the same everywhere don’t actually, so there are workarounds in many places. (But you don’t have to worry about them _using_ DDEV!)
```

Is it too reductive to equate ARM64 and Apple Silicon? A Raspberry Pi isn’t Apple Silicon but it is ARM64, right?

src/content/blog/docker-providers.md
```diff
@@ -0,0 +1,58 @@
...
+| Windows WSL2 | docker-ce | yes | yes | Standard Linux Docker, performant, free, open source, well-maintained              |
+| Windows WSL2 | Docker Desktop | no | no | Can be buggy                                                                       |
+| Linux | docker-ce | yes | yes | Standard open-source Docker, well-maintained                                       |
+| Linux | Docker Desktop for Linux (DO NOT USE) | no | no | Non-functional for containers like DDEV that don’t use root user, poor performance |
+
+## What are DDEV Users Choosing Right Now?
+
+Traditional Windows users have only the Docker Desktop option, and nearly everyone on Linux uses docker-ce, but here are the macOS and WSL2 breakdowns:
+
+![Macos Docker Platform Usage](/img/blog/2023/11/macOS_Docker_Platform.png)
```
mattstein, line 56: The alt text here will describe the image for someone that may not be able to see it, so it might be nice to summarize what’s happening in the chart. (I’d use percentages but I don’t know what they are.)

```suggestion
![Donut chart depicting the last 7 days of macOS Docker Platform usage, led by almost three quarters Docker Desktop, then Colima, and a sliver of OrbStack](/img/blog/2023/11/macOS_Docker_Platform.png)
```

src/content/blog/docker-providers.md
```diff
@@ -0,0 +1,58 @@
...
+
+![WSL2_Docker_Platform Usage](/img/blog/2023/11/WSL2_Docker_Platform.png)
```
mattstein, line 58: Same deal: 

```suggestion
![Donut chart depicting the last day of WSL2 Docker Platform usage, with slightly more than half using WSL2 and slightly less than half using Docker Desktop](/img/blog/2023/11/WSL2_Docker_Platform.png)
```

src/content/blog/docker-providers.md
rfay: It might be. We test on linux arm64, of course, and a RPI is the main gateway to the buildkite test runners. But for most people, Apple Silicon == ARM64.

public/img/blog/2023/11/docker-in-a-box.png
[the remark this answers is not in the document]
mattstein: If it’s a smaller file size and looks just as good to you there’s no need to use my old-fashioned, hand-spun JPEG.

stasadev (approved):
Looks good to me 👍

PR #150, opened 2023-11-29
Add Astro v3 ViewTransitions support.

bmartinez287:
Adds View Transitions support.
https://www.youtube.com/watch?v=4Ohi9G-RCcE&ab_channel=LearnWithJason
I saw this on the video above the other day.


bmartinez287:
yeah, that's one tricky bug to overcome.

rfay:
Does astro v4 help with any of this?

bmartinez287:
Not sure it does. As this is a bit of an issue between native javascript API's.
The transition API makes some assumptions that our current script does not take into account.

PR #151, opened 2023-11-29
Blog - Show Docker provider performance

rfay:
## The Issue

Blog update about Docker provider performance


micahthomas:
@rfay 

I wanted to check w/ you on the colima command used to test vz.
Is the command `colima start vztest --cpu=4 --memory=6 --disk=100 --mount-type=virtiofs`?

I was trying to replicate the results using orbstack vs colima (vz) vs colima (qemu), and I am not seeing your results. I suspect the reason is if you use the above command, colima actually does not use virtiofs/vz. You also have to pass in the `--vm-type=vz` option for it to properly use virtiofs.

Based on my tests for installing ddev using `ddev start` on a clean colima/orbstack instances:

colima qemu:
```bash
79.56 real         1.34 user         1.05 sys
```

colima vz (virtiofs):
```bash
69.31 real         1.38 user         1.06 sys
```

orbstack:
```bash
68.75 real         1.48 user         1.19 sys
```

rfay:
I don't think that you can do virtiofs without vz on colima so I imagine I had it right. And isn't vz the default now?

Anyway, would love to have a followup blog post from you with your results and details and approach!

PR #160, opened 2024-02-19
Blog - Setting up a go development environment

bmartinez287:
DDEV Contributor Live Training: Setting up a Go environment blog version.


rfay (commented):
Awesome, thanks! I'll have some comments and fixes, but thanks so much for doing this!

bmartinez287:
Highlighting the Live Training is tricky since a lot of sections have a lot of context provided on the video. 
In this blog post, I tried to highlight the timeframe in which those happened and important subjects covered on it.
Its flow as an article could use work but on the plus side it is a time saver when looking for specific DDEV knowledge. That said,  we can updated as needed.

rfay:
Put on my list

rfay (commented):
Do you mind if I push a commit or two into this? It's really useful, but I'd like to use a slightly different approach in a couple of places for clarity. It's hard to do what I'd like to do just with the commenting approach.

rfay:
Please take a look and see what you think. I spent quite a bit of time on it but perhaps not enough. Moved the video to youtube so it's easy to link into the exact places.

bmartinez287:
oooh, that's cool.
I wonder if we should do the blog and also annotate the video with that information. 
Adding timestamps to a YouTube video makes it easier to parse the information.
It's nice to see it on the video itself.

rfay:
I don't know how to do it and don't have much more bandwidth for this one, not sure if I could give you permissions.

I wish we still had control of the `ddev` user in there, was lost with the drud company.

src/content/blog/setting-up-a-go-development-environment
```diff
@@ -0,0 +1,92 @@
+---
+title: "Setting up a Go Development Environment"
```
rfay, line 2:
```suggestion
title: "Setting up a Go Development Environment for DDEV"
```

rfay:
Thanks for the careful work on this! In the future it would definitely be best for each contributor training to have a blog to accompany it in the beginning.

bmartinez287:
In the beginning as in before it comes out? Or closer to when it comes out?

rfay:
It would be better if I wrote a blog to go with the training :) And of course updated relevant docs. But that might be putting too much pressure on myself. Doing these informally worked out pretty well last year.

bmartinez287:
Yeah, that be nice. But the recording accomplishes 80% of the job. I would rather have more recordings with no blogpost than less of them that have it and are more structured.
Saving energy on some tasks is important.

Plus, videos usually by their nature include extra info as people ask questions or share insights that cover new areas.

PR #164, opened 2024-03-01
2023 review: Add more about contributor training

rfay:
## The Issue

@bmartinez287 pointed out the importance of contributor training in 2023.


bmartinez287:
It be great to hear more about the contributions Simon made. Maybe a highlight of the biggest items specially since we are mentioning Matt.

bmartinez287 (approved):

rfay:
> It would be great to hear more about the contributions Simon made

I could write a whole blog post just about that. I think it would be a distraction in the midst of this post though. I wish he were here so we could honor him with a post like that.

PR #170, opened 2024-03-14
Updated and simplified ddev config.yml

bmartinez287:
## The Issue
DDEV setup commands where to complex

Simplifies the workflow and provides a better experience out of the box.


.ddev/config.yaml
```diff
@@ -20,14 +19,17 @@ web_extra_exposed_ports:
...
 web_extra_daemons:
   - name: astro-dev-daemon
-    command: "sleep 3 && npm run dev -- --host" # sleep avoids race conditions
+    command: bash -c 'npm install && touch /var/tmp/npminstalldone && npm run dev -- --host'
     directory: /var/www/html
 hooks:
   post-start:
-    - exec-host: echo -e "                                  NOTICE                                        \n
+    - exec: bash -c 'while [ ! -f /var/tmp/npminstalldone ]; do sleep 1; done'
+    - exec: npm run build
```
rfay, line 27: Why isn't this in the daemon startup instead of here? Another `&&` ?

rfay (commented):
Looking good! When you like it I like it.

.ddev/config.yaml
bmartinez287: great question, I was going to add it there but npm run build takes 30seconds.
Screenshot 2024-03-14 at 2 24 05 PM
and since it on that background daemon it creates a lag plus we only need it the first time around since most times we be running HMR(hot module reloading anyway).

How do I turn the echo -e green? I wanted ${DDEV_PRIMARY_URL}:4321 that to be green to but i can't find the variable.

rfay: Pretty fancy, try https://stackoverflow.com/questions/5947742/how-to-change-the-output-color-of-echo-in-linux

bmartinez287: `echo -e "\e[1;31mHello \e[1;35mWorld"`
that works locally but once it goes through ddev exec it fails.
ddev exec echo -e '\e[1;31mHello \e[1;35mWorld'
e[1
bash: line 1: 31mHello: command not found
Failed to execute command echo -e \e[1;31mHello \e[1;35mWorld: exit status 127

rfay: Try it in single quotes.

bmartinez287: Screenshot 2024-03-14 at 4 13 19 PM
Got it. I must have been having a typo there.

bmartinez287:
It should be good to go now.

rfay:
You have some rebasing that needs to be done.

bmartinez287:
config has been rebased.

rfay (approved):
Thanks!

PR #173, opened 2024-03-15
Add npm run build to the ci test.

bmartinez287:
## The Issue
The build might fail crashing the site.

Add npm run build to the ci test.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,3 +18,5 @@ jobs:
         run: npm ci
       - name: Lint
         run: npm run textlint
+      - name: Lint
+        run: npm run build
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,6 +10,9 @@ on:
 permissions:
   contents: read
 
+env:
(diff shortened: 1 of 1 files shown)


rfay (approved):
Awesome, thanks!

rfay (changes requested):
I guess we'll have to provide a token to make this work. Or figure out how to skip all the image stuff. Or just do the `npm run dev` ?

rfay:
This only takes a token that has no power at all. There should be a way to do it!

rfay:
I think we should just be able to just provide the $GITHUB_TOKEN from the *forked* PR to the workflow and it would be adequate for this. Not sure.

rfay:
I think this just worked!

rfay (approved):
Seems like we have it here! If you like it pull it.

rfay:
This may still fail with forked PRs. Let's get it in and then we'll see.

bmartinez287:
that's cool, I imagine you used an organizational token instead of a personal one.
Lets try it.

PR #179, opened 2024-03-28
switch default to Orbstack for get started

rfay:
## The Issue

Our default/easiest recommendation these days is Orbstack, not Docker Desktop

Switch to the new recommendation

Review at https://20240328-use-orbstack.ddev-com-front-end.pages.dev/get-started/ - macOS version

### DIFF
--- a/src/components/quickstart/Mac.astro
+++ b/src/components/quickstart/Mac.astro
@@ -12,7 +12,7 @@ const { latestVersion } = Astro.props
 
 <div class="prose dark:prose-invert">
   <p class="text-gray-600 mb-16 dark:text-slate-400">
-    We’ll use Docker Desktop since it’s the most straightforward of <a
+    We’ll use Orbstack since it’s the most straightforward of <a
       href="https://ddev.readthedocs.io/en/stable/users/install/docker-installation/#macos"
       >several options</a
     >.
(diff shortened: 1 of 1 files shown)


src/components/quickstart/Mac.astro
```diff
@@ -48,12 +47,11 @@ const { latestVersion } = Astro.props
 
 <div class="prose dark:prose-invert">
   <p>
-    Launch Docker Desktop, accept the license agreement, and make sure Docker is
-    started and running.
+    Launch Orbstack, accept the license agreement.
```
rpkoller, line 50: the only nitpick in the original there is the part about "and make sure docker is started and running" . not sure about the details of orbstack but is the vm started when orbstack is launched automatically or has the user still to start it after starting orbstack? in case that would apply to orbstack i would reintroduce the striked last part about started and running?

rpkoller: aside that it looks good

stasadev (approved):
Looks good to me 👍

src/components/quickstart/Mac.astro
rfay: > but is the vm started when orbstack is launched automatically or has the user still to start it after starting orbstack

Yes, it is started automatically.

PR #180, opened 2024-03-28
Improve ddev-name-resolution-wildcards.md with reference to debugging

skeemer:
I think this is a better test as it won't likely be in a `hosts` file and covers the situation where first level wildcards work and deeper fails. It also answers the question directly and provides a next step.


rfay:
Thanks!

rfay:
I went with `test.ddev.site` after all, simpler. Added a link to the question.

PR #181, opened 2024-03-29
Updates ddev.com to astro 4.5, fixes #165, fixes #156, fixes #155, fixes #157, fixes #159

bmartinez287:
## The Issue
Updates Astro to 4.5 and it's dependencies.

The main issues here were:

1. Figure out why glob.sync won't work and if it does not find a replacement for glob.
2. astro-imagetools did not work with the latest version of Astro so we had to switch back to their native Images package which I believe was not released when we went for imagetools.

Glob does the website search and it checks the author's name against the MD name field.


bmartinez287:
Rss and json feed are also broken and need to be updated.

rfay:
Just a note to say that your maintenance of this is very much appreciated!

rfay:
Will this PR obsolete all the pending dependabot dependency PRs?

bmartinez287:
Yeah, This should take care of all those updates/warnings.
I started with the largest packaged and will work my way down.

bmartinez287:
This should be ready for review.
It's unfortunate that I could not find an updated version of the Astro-image tools. But we can use Astro native images and pictures even if it does not have all the additional options.

Upcoming changes:

- Astro-image is removed since they don't have a compatible release and we move towards using Astro native image and picture component.
- base64-img has all kinds of security warnings on its dependencies and the latest release was 6 years ago hence we move towards node-base64-image which seems to do the same. This one package had 7 issues that drove dependabot crazy.
- the glob package import was updated so search and other elements that use it should work as expected.

bmartinez287:
Closes: #165, #156, #155, #157, #159

rfay:
> It's unfortunate that I could not find an updated version of the Astro-image tools. But we can use Astro native images and pictures even if it does not have all the additional options.

Could you say more about what this means? Are we talking about the tools the build the site? Or testing tools?

bmartinez287:
Neither

rfay:
Rebased.

rfay:
🚀 🥇

PR #182, opened 2024-03-29
Update Discord link

stasadev:
## The Issue

- https://github.com/ddev/ddev/pull/5637

Uses the new link.


rfay (approved):
This is correct, thanks!

PR #195, opened 2024-04-23
fix: make tests work again, bump GitHub dependencies

stasadev:
## The Issue

Tests are broken:

```
Warning: No existing directories found containing cache-dependency-path="./yarn.lock"
```

I investigated this and found that it is a problem in `withastro/action@v2`

I opened a PR there, but don't know if it is going to be merged soon:

- https://github.com/withastro/action/pull/52

1. Removes all `yarn.lock` files from `node_modules` directory before running `withastro/action@v2`. (This workaround can be removed later.)
2. Bumps `actions/checkout` and `actions/setup-node` to `v4`.
3. I found out that `withastro/action@v2` uses `node-version`, but we do not set it explicitly - fixed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,7 @@ jobs:
         run: npm ci
       - name: Lint
         run: npm run textlint
+      # Remove this when https://github.com/withastro/action/pull/52 is merged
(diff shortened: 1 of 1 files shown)


rfay (approved):
Awesome, thanks!

PR #199, opened 2024-05-03
docs: install curl, update codespaces link

stasadev:
## The Issue

- https://github.com/ddev/ddev/issues/5915

Adds `curl` installation for Debian/Ubuntu.
Updates the link for GitHub Codespaces.


rfay:
Thank you so much for remembering this!

It might be better just putting curl in the earlier instruction, 
image

stasadev:
Okay, I moved it to the top.

PR #209, opened 2024-06-05
test: remove workaround for withastro/action, for #195

stasadev:
## The Issue

- #195

My workaround is no longer needed as the fix has been merged upstream.
Checking if the tests work without it.


rfay:
Pull when you're happy with it.

PR #223, opened 2024-07-14
Remove extra slash in the search URL results

stasadev:
## The Issue

When you search something here https://ddev.com/search/, it produces a link with double slash

image

Removes this extra slash.

Try searching here https://20240714-stasadev-search-ext.ddev-com-front-end.pages.dev/search/


rfay (approved):
Nice work, thanks!

PR #224, opened 2024-07-14
Use const for address on DDEV Foundation page

stasadev:
## The Issue

https://ddev.com/foundation/ could read the org address from `const`.

Conversely, this PR replaces this one, because you can't use JS variables in markdown:

- #76

Uses `const` variables.

Updates postal code for DDEV Foundation.

Review https://20240714-stasadev-foundation.ddev-com-front-end.pages.dev/foundation/


rfay (approved):
Thanks!

PR #238, opened 2024-08-09
Backups with DDEV blog

rfay:
## The Issue

Blog on backups with DDEV

Rendered version at https://20240406-backups.ddev-com-front-end.pages.dev/blog/ddev-backups/


src/content/blog/ddev-backups.md
```diff
@@ -0,0 +1,55 @@
...
+If you prefer a text-based database backup, you could do this:
+```yaml
+hooks:
+  pre-stop:
+    - exec-host: 'mkdir -p .tarballs && ddev export-db --file=.tarballs/db.$(date +"%Y%m%d%H%M%S").sql.gz'
+```
+
+**PHP Code and Configuration** are text files that should be under control of your Git environment. The easy answer for those is to always keep them pushed up to your Git provider. Work on a branch and a new commit up to your branch regularly, so you could start afresh on any computer any time. This is a wonderful thing about source control!
+
+**User-generated and non-code images, etc.**: In general, these should be coming from upstream hosting environments. Although it is possible to push them upstream, it's usually a bad idea, as it could overwrite a production environment. So normally if you lost these, you would just pull them again from wherever they come from.
```
stasadev, line 53:
```suggestion
**User-generated and non-code images, etc.**: In general, these should come from upstream hosting environments. Although it is possible to push them upstream, it's usually a bad idea, as this could overwrite a production environment. So normally, if you lose these, you would just pull them again from wherever they come from.
```

src/content/blog/ddev-backups.md
```diff
@@ -0,0 +1,55 @@
...
+* Configuration
+
+We need a backup strategy for each of these. 
+
+**Choose a trusted local backup solution.** Time Machine on macOS can work wonderfully for many people, and there are lots of other solutions out there. However, databases have not traditionally been good with backups because they are often not consistent during the backup process, so we need to do something different with them.
+
+**But try not to rely on the local computer anyway.** I try to push code and config to my Git provider (GitHub, GitLab, etc.) and make sure I know how to recreate any user-generated files. Make sure your database contents are accessible from upstream sources like production, or keep good snapshots or exports.
+
+**Databases** need an extra step for safety. In general, don't do work on a database that cannot be recreated by code (with migrations, for example). But most of us want to have a database quickly available as a good place to start from. Since databases are typically in a binary format that can't reliably be backed up, we need a good way to make a copy. DDEV has two great ways to turn databases into files, `ddev snapshot` and `ddev export-db`.
```
stasadev, line 30:
```suggestion
**Databases** need an extra step for safety. In general, avoid working on a database that cannot be recreated by code (with migrations, for example). However, most of us want to have a database quickly available as a good place to start from. Since databases are typically in a binary format that can't reliably be backed up, we need a good way to make a copy. DDEV has two great ways to turn databases into files, `ddev snapshot` and `ddev export-db`.
```

stasadev (approved):
Looks great, minor suggestions to make it more formal.

PR #240, opened 2024-09-02
`mkcert -install` should be shown in ddev.com quickstarts

rfay:
## The Issue

@philalonso [pointed out in Drupal slack](https://drupal.slack.com/archives/C5TQRQZRR/p1725301877083119) that he had done a DDEV installation and ended up with no https 

He checked carefully and found that they'd used https://ddev.com.ddev.site/get-started - and in fact it does not show the `mkcert -install` step.

* Add the `mkcert -install` to macOS and Linux. (It's done automatically by the WSL install scripts)
* I noted that the Debian install used the older instructions for the key location, so fixed that.


stasadev (approved):
I did this:
- added one more commit to remove duplicate code from Get Started
- reviewed Get Started tabs and updated little things here and there

rfay:
Thanks for fixing Cloud section. I ran out of gas on that :)

Still not confident about Codespaces of course, but what you have is better.

Thanks for the examples reorg.

rfay:
The one thing that Matt would be saying is that Linux/Cloud should be bash with $ and mac with the arrows. But I'm OK with it.

PR #241, opened 2024-09-03
GitHub style for inline code blocks

stasadev:
## The Issue

I have a hard time telling if I'm looking at the code or something else.

Use style similar to GitHub.

Example from https://ddev.com/blog/anatomy-advanced-ddev-addon/

Before:
image

After:
image

Before:
image

After:
image

### DIFF
--- a/tailwind.config.cjs
+++ b/tailwind.config.cjs
@@ -36,9 +36,6 @@ module.exports = {
             // remove automatic quote marks from blockquotes
             "blockquote p:first-of-type::before": null,
             "blockquote p:last-of-type::after": null,
-            // remove automatic backticks that surround inline code blocks
-            "code::before": null,
-            "code::after": null,
           },
           code: {},
         },
--- a/tailwind.config.cjs
+++ b/tailwind.config.cjs
@@ -29,6 +29,10 @@ module.exports = {
           800: "#001f38",
           900: "#00192d",
         },
+        code: {
+          light: "#818b981f",
+          dark: "#2e3440ff",
+        }
       },
       typography: {
         DEFAULT: {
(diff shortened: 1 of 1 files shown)


rfay:
You say 
> Bring back the backticks

Does that mean there is a history to this?

stasadev:
See the change, Tailwind CSS adds backticks by default to the `<code></code>` blocks.
And these backticks were removed while developing this website.

rfay:
I can see why it was done, as the backticks are "literal", whereas in markdown we would never see them. With this change, we see them. As with markdown, backticks are part of the source, not part of the presentation.

image

stasadev:
The reason why I want to bring them back is that the markdown command is not always clearly visible, it just blends in with the rest of the text (I can see it's just bolder text and that's it).

I just want to emphasize these blocks of code somehow, not necessarily with backticks, another idea would be to add a background like here on GitHub.

rfay:
That's completely opposite how markdown works, which is where the backticks idea comes from. In markdown the *source* has backticks, but the presentation does not, instead it has styled text.

I think perhaps what you want is different styling for these presentations? Maybe you like the github markdown styling better?

stasadev:
> I think perhaps what you want is different styling for these presentations? Maybe you like the github markdown styling better?

Yes, that's what I want.

I investigated Tailwind config, and made necessary changes.
Please check the before/after in the first post.

rfay:
I sure like how it looks! https://20240903-stasadev-code-backt.ddev-com-front-end.pages.dev/blog/anatomy-advanced-ddev-addon/

rfay (approved):
Looks fantastic to me, thanks!

PR #243, opened 2024-09-07
[TASK] Add blog post about using DDEV in GitLab CI

ochorocho:
## The Issue

> People have been asking about this for years and it would be lovely for them to know about it. Anywhere would be great, 
> but how about a guest blog on ddev.com? It's just markdown and super easy, 

https://github.com/ochorocho/ddev-gitlab-ci/issues/11


rfay:
Thanks!

rfay (changes requested):
This is going to be great, thanks! 

* The featuredImage isn't working right:

image

* Please make sure to test and read it. It's just a `ddev start` to review it so you can see things like that and see how it renders.
* Please make a strong introduction instead of talking about whatever those glorious days were :)  The intro should say something strong like "Now you can use DDEV for CI on Gitlab" and what that means. I think it's fair to assume your readers have already bought into the idea of using CI.
* Please add a link to a public working workflow on gitlab if that's possible
* Please add a link to the equivalent (I think?) GitHub Actions action, https://github.com/ddev/github-action-setup-ddev

Thanks!

ochorocho:
Well, the featuredImage was ignored by my global .gitignore 🤦 it should be displayed now.
Added/changed the text according to your suggestions.

Thank you.

rfay (commented):
It's looking great, but for a n00b there's more that can be done both here and in the README, so some comments. (This could go in as is, but I'd love to talk about these things...)

* Could you just say a bit about how testing is done or enabled in gitlab? Since I've never done that, I'm kind of lost at the very beginning here (and also in the project README). 
* This and the README can be dramatically improved with a specific example. (I spent some time wondering if the `build.sh` was important to me, it obviously isn't. But it's the most prominent thing in the README.) So an example might be... "If you have a TYPO3 project checked into Gitlab and want to add testing that uses DDEV, ..."
* Shouldn't the default be current DDEV stable version? Why all the references to really old DDEV versions, and why should one have to specify a version?
* Important projects sometimes have a hard time living on in a personal repository like ochorocho. I invite you to move this to the ddev org if you're open to that. Alternately, move it into a TYPO3 org or b13 ? If you're interested in moving it, it could move sooner rather than later and then the article wouldn't need updating at that time.

Interesting factoid: I disabled automated tests on `no-bind-mounts` quite some time ago because I'd never heard of anybody using it... And the only person that I remember having used it on gitlab failed because of the lack of write permission on the remote docker server. That was @moshe, but I think he was also using the [remote docker instance](https://ddev.readthedocs.io/en/stable/users/topics/remote-docker/#remote-docker-instances), I guess you've solved that problem by using DinD, which wasn't as mature at that time.

ochorocho:
> Could you just say a bit about how testing is done or enabled in gitlab? Since I've never done that, I'm kind of lost at the very beginning here (and also in the project README).

Which testing are your talking about? The bats tests for the container or actual e2e (end to end) tests?
Well, i'm not sure if it makes sense to explain how to install ddev addons and set up playwright.
It feels a bit out of scope.

> This and the README can be dramatically improved with a specific example. (I spent some time wondering if the build.sh was important to me, it obviously isn't. But it's the most prominent thing in the README.) So an example might be... "If you have a TYPO3 project checked into Gitlab and want to add testing that uses DDEV, ..."

I've extended and restructured the README to be more clear what the important sections are.

> Important projects sometimes have a hard time living on in a personal repository like ochorocho. I invite you to move this to the ddev org if you're open to that. Alternately, move it into a TYPO3 org or b13 ? If you're interested in moving it, it could move sooner rather than later and then the article wouldn't need updating at that time.

I'm happy to move it to the ddev org.

ochorocho:
FYI, i've moved the project to the ddev org :-)

ochorocho:
Sorry forgot to answer this one:
> Shouldn't the default be current DDEV stable version? Why all the references to really old DDEV versions, and why should one have to specify a version?

So you're saying the image should have a `latest` tag too?

rfay:
You sure move fast!!!! Thanks!

> Which testing are your talking about? The bats tests for the container or actual e2e (end to end) tests?

I was talking about using DDEV in other tests, not the tests for this add-on. People like me would just like to understand in context how this will be used.

> So you're saying the image should have a latest tag too?

No, I'm probably saying that it should automatically use the current stable, unless another version is specified.

This is all so great, I'll follow up with issues in the repo, will take one more look at the article here and we'll get it moving along. I imagine you already changed the article to change the URL and such.

ochorocho:
> I imagine you already changed the article to change the URL and such.

Not just yet. Will do later today or early tomorrow. :-)

rfay:
Instead of going back-and-forth I pushed one small commit and will pull it now. Thanks so much for this contribution and for maintaining it!

ochorocho:
@rfay thanks for all the support you provided on the way!

PR #257, opened 2024-10-11
Update animated terminal

stasadev:
## The Issue

The output is slightly different.

Updates it.

Look at the animated terminal here https://8b16f7c7.ddev-com-front-end.pages.dev/


rfay (approved):
So great that you are paying attention to this, thanks. Pull whenever you're ready.

PR #259, opened 2024-10-11
Change to use short link and 301

rfay:
## The Issue

@stasadev pointed out I should use /s prefix and 301 redirect. Done here, updated README


stasadev (approved):
Thank you!

PR #265, opened 2024-10-25
Improve README for new contributors

mandrasch:
## The Issue

- Not clear that you need to open https://ddev.com.ddev.site:4321/ in your browser

- add info
- add info for troubleshooting

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,10 +66,11 @@ All commands are run from the root of the project, from a terminal:
 DDEV already has all the dependencies included.
 
 1. Run `ddev start` to start and set up the project’s dependencies.
+2. Open https://<projectname>.ddev.site:4321 in your browser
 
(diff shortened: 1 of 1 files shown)


rfay (approved):
Looks good to me, thanks! It does prompt you on `ddev start` to know what URL to use, and it also shows it in `ddev describe`. I usually end up using `ddev describe` a few times. We should probably add an extra launcher custom command or something.

mandrasch:
> Looks good to me, thanks! It does prompt you on `ddev start` to know what URL to use, and it also shows it in `ddev describe`. I usually end up using `ddev describe` a few times. We should probably add an extra launcher custom command or something.

Sure thing 👍

Yeah, I saw the prompt ... but then switched terminals and used `ddev launch` because I forgot --> this only renders `dist/` (also a bit confusing maybe, why a page is suddenly showing up there)

From my personal experience, `ddev describe` is often unknown or people forget about it (unfortunately). Always trying to educate them of course. :-D

Issue #266, opened 2024-10-25
Make GITHUB_TOKEN optional for local development?


mandrasch:
In my personal opinion it is a bit of a challenge that people who want to add a blog article PR need to setup a GitHub key first. 

They also might not understand why they have to do it on their own account. An information what the key does is also missing in README.md

And ... first it works without a token, but then you run into the GitHub API rate limit errors when you try `ddev npm run build`. 

My proposal would be something like this for `api.ts`:

```js
export async function getSponsors() {

  if(!process.env.hasOwnProperty('GITHUB_TOKEN') || process.env.GITHUB_TOKEN === ''){
    return [];
  }
```

If there is not `GITHUB_TOKEN`, we simply don't render sponsors, contributors, latest releases.

We might add a warning if this is a production build and the `GITHUB_TOKEN` is still missing.

(Goal: Make it as easy as possible to contribute, guess adding blog articles via Gitpod would also be great)

rfay:
Wow, I never set GITHUB_TOKEN. I just `ddev start` and move on. Maybe I have something set that I don't understand?

rfay:
I guess it's because I don't ever rebuild those, and I guess that must not be required.

rfay:
It would be a fine addition, thanks!

mandrasch:
You have this cached in `cache/` most probably already, there is a nice caching mechanism in api.ts:

```
const cacheFilename = 'releases.json'
const cachedData = getCache(cacheFilename);
```

When I cleared my `cache/`-folder today, and ran `ddev npm run build` today - I had this:

```
16:19:39 ▶ src/pages/about.astro
16:19:39   └─ /about/index.htmlRequest quota exhausted for request POST /graphql
```

rfay:
My first time up on a new machine I also got this:

> 08:22:41   └─ /about/index.htmlRequest quota exhausted for request POST /graphql

And it also seems to hang forever there.

rfay:
I imagine I had a .env on other machine that I wasn't aware of. 

Note that it seems to work with a no-scope-no-privileges token (I don't think it needs the privs currently stated in the README)

mandrasch:
Created draft https://github.com/ddev/ddev.com/pull/268

mandrasch:
> Note that it seems to work with a no-scope-no-privileges token (I don't think it needs the privs currently stated in the README)

Ah interesting! Briefly wondered why these are needed when you are someone with no privs for github.com/ddev anyhow

[...]

PR #267, opened 2024-10-25
Make GitHub token optional

mandrasch:
## The Issue

https://github.com/ddev/ddev.com/issues/266

- add check in the beginning of api.ts, IIFE (Immediately Invoked Function Expression) pattern
- add checks to all functions calling github API, return empty arrays

- test with and without GITHUB_TOKEN, local development on https://ddev.com.ddev.site:4321/ as well as `ddev npm run build`

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,10 +57,6 @@ All commands are run from the root of the project, from a terminal:
 
 ### Local Development Setup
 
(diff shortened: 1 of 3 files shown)


mandrasch:
branch was wrong, need to re-do

PR #276, opened 2024-11-12
Makes the hero wider so it does not look tiny on larger screens and other minor fixes

bmartinez287:
It makes the hero wider so it does not look tiny on larger screens, like the footer, and provides more space to the hero.

## The Issue

Before this PR
Screenshot 2024-11-11 at 7 02 32 PM

After this PR.
Screenshot 2024-11-11 at 7 02 15 PM

This is mostly to improve how the site looks on larger screens.

Review at https://20241011-bmartinez-header-fo.ddev-com-front-end.pages.dev/

### DIFF
--- a/src/components/Footer.astro
+++ b/src/components/Footer.astro
@@ -56,7 +56,7 @@ const navigation = {
 ---
 
 <footer class="dark:bg-transparent">
-  <div class="max-w-4xl mx-auto py-12 px-6 lg:px-0 lg:py-16">
+  <div class="max-w-7xl mx-auto p-8">
     <div class="pb-8 xl:grid xl:grid-cols-3 xl:gap-8">
       <div class="grid grid-cols-auto gap-8 xl:col-span-4">
         <div class="md:grid md:grid-cols-4 md:gap-8">
--- a/src/components/Header.astro
+++ b/src/components/Header.astro
@@ -64,7 +64,7 @@ import {
 
 <header class="w-full p-8">
   <nav
-    class="flex justify-between items-center content-center max-w-4xl mx-auto"
+    class="flex justify-between items-center content-center"
     aria-label="Main"
   >
     <a href="/" class="relative block cursor-pointer -ml-2">
--- a/src/components/Hero.astro
+++ b/src/components/Hero.astro
@@ -10,7 +10,7 @@ import { CloudIcon } from "@heroicons/react/24/solid/index.js"
 ---
 
 <div class="mt-2 pt-8 sm:mt-12 sm:pt-24 pb-32 relative overflow-x-hidden">
-  <div class="z-10 max-w-4xl mx-auto relative">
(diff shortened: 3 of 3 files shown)


rfay:
What is a hero? You're a hero!

rfay:
As I make the width less, it gets ugly, 

image

bmartinez287:
That's an interesting observation.
However, that seems like the same to me compared to ddev.com.
As far I remember that was the original design for it.
The terminal appears only about 2500px.
Screenshot 2024-11-12 at 9 58 51 AM

This PR mostly switches the brandbar align the items left and right and the footer.

bmartinez287:
I'm thinking about changing the hero layout but that likely be a different PR.

rfay:
I note that it's pretty terrible on phone.

Thanks for paying attention, sorry for the distraction!

stasadev (commented):
I have an ultra-wide display and I don't like this change (sorry).

Before:
before

After:
after

It's too much for me.

For comparison what is used for the news websites:

It's easier to see in the center, and I don't have to move the mouse too much.

bbc

cnn

nytimes

stasadev:
It's 2560x1080.

stasadev:
I think Tailwind has a different layout, because the logo is not clickable (it's one page site).
And for DDEV we want to have a way to go to the main page from a blog post page.
And with this change, I need to move the mouse far to the left.

rfay:
I guess we should close this if it doesn't have a future?

bmartinez287:
Do we know what inspired the current design? Was there Figma or something like it involved or not?

I might close this issue and take another angle at it. Initially, I was going to make the changes incrementally. However, that is challenging as the current design makes some assumptions.

I could start with the interior pages that need work, but we could revamp the homepage and make that carry throughout. Plus with Astro 5 released 2 days ago and tailwind4 coming in it be nice for a refresh.

thoughts? interior pages or homepage redesign?

rfay:
@mattstein knows :)

mattstein:
Not sure how much help it is, but I cobbled things together roughly looking at Tailwind UI and trying to approximate the [previous version of the site](http://web.archive.org/web/20220804135858/https://ddev.com/) without any radical departures. Nothing ever touched Figma because I just started noodling with it wondering what an update might look like and not originally intending to run with it.

My aim was to keep a similar, solid column that the previous site used and I only broke out of it slightly on the homepage for a bit more rhythm and breathing room. The hero graphic/animation was meant to peek out from the side and give a whiff of what DDEV looks like in action rather than be a prominent feature, but that admittedly gets awkward with a wide viewport.

No strong feelings about what this PR is doing, I only wonder what the implications are for the rest of the site since it does conform more strictly to a single, tall column.

bmartinez287:
Thanks for the feedback Matt! I can relate, "just started noodling" been there done that haha.

rfay:
If this has a future feel free to reopen or try again, closing for now.

PR #285, opened 2024-12-05
Auto-format files with prettier, add vscode settings

mandrasch:
Moved to https://github.com/ddev/ddev.com/pull/287


mandrasch:
Here is the promised PR - I tested some blog posts locally and they all seemed fine. Don't know if it is necessary to run a visual comparison test between locahost:4321/ and https://ddev.com to catch differences?

rfay:
Thanks so much @mandrasch ! 

I apologize, I forgot that only PRs on the upstream branch get preview builds. I've invited you to be able to do that on ddev/ddev.com, would you mind pushing this there and creating a new PR so it would be easy to do the visual review?

mandrasch:
Moved to https://github.com/ddev/ddev.com/pull/287

PR #288, opened 2024-12-06
Add discord shortlink

rfay:
## The Issue

@stasadev pointed out that we could have a discord shortlink instead of the obscure link we currently use, https://discord.com/invite/5wjP76mBJD

Use https://ddev.com/s/discord and redirect.

Test at https://20241206-discord-shortlink.ddev-com-front-end.pages.dev/s/discord

Followup would be to update ddev.com, docs, and any other references.


stasadev (approved):
Thank you, it looks good.

And it's so much easier to share.

stasadev:
Tested on Android mobile, the Discord app opens as expected.

PR #293, opened 2024-12-19
Upgrades Astro to 5.1 and its dependencies

bmartinez287:
## The Issue

Atro 5.1 release dropped this week and it requires a few minor changes on the architecture side.
Most notably there's no slug property is called ID now.


bmartinez287:
Screenshot 2024-12-19 at 3 01 43 PM
So far I'm mostly noticing one bug at compile time. 
that information on the red square needs to be loaded in a slightly different way.

bmartinez287:
I'm not sure why Cloudflare crashed at the beginning.

rfay:
> 16:39:37.413 | Node.js v18.16.0 is not supported by Astro!
> 16:39:37.413 | Please upgrade Node.js to a supported version: ">=18.17.1"


I pushed an update to Ubuntu 24.04 for the test runner, but we may need to do something with that as well.

rfay:
Changing the .nvmrc to v20 seems to have fixed Cloudflare deployment.

stasadev:
> Screenshot 2024-12-19 at 3 01 43 PM So far I'm mostly noticing one bug at compile time. that information on the red square needs to be loaded in a slightly different way.

Fixed it with `await` [`324cdb7` (#293)](https://github.com/ddev/ddev.com/pull/293/commits/324cdb707a7f5f079edf14e14490ab8d7af7efc3)

stasadev (approved):
Looks good to me! Thank you.

I had to delete the project locally using `ddev delete -Oy`, as HMR wasn't rendering properly. (It might have been a caching issue on my end.)

bmartinez287:
so I commented out await because it would fail there initially, but it might have been a local glitch of mine because it works now. Yay.

PR #298, opened 2024-12-24
Add /contact page, use instead of email

rfay:
## The Issue

* I noticed that the `email` in the footer didn't actually work, at least on my browser
* The expected /contact page didn't exist, most org sites want that.
* Add LinkedIn to footer socials
* Minor improvement to layout of snailmail contact

Rendered at https://20241224-contact-and-linkedi.ddev-com-front-end.pages.dev/

See especially the footer and /contact pages


src/const.ts
```diff
@@ -11,17 +12,18 @@ export const GITHUB_ISSUES_URL = "https://github.com/ddev/ddev/issues"
 export const DOCUMENTATION_URL = "https://ddev.readthedocs.io/"
 export const DISCORD_URL = "/s/discord"
 export const EMAIL_URL = "mailto:support%40ddev.com"
```
bmartinez287, line 14: This typo to be specific we need to replace %40 for @

bmartinez287:
Screenshot 2024-12-25 at 9 12 54 PM
the email issue seems to carry the on both places.
It looks like it was a typo on that variable. i added a comment to that.

bmartinez287:
other than that the contact page seems fine.

src/const.ts
rfay: This works in regular links, and the idea is to prevent web crawlers from capturing the email address.

bmartinez287: oooh, that makes sense, I thought this was the original issue with the email on the footer.

bmartinez287 (approved):

stasadev (approved):
I added LinkedIn SVG and fixed this error in the dev console
image

Looks good to me!

rfay:
You're awesome, thanks!

PR #329, opened 2025-02-19
Prototypes pricing page for ddev.com

bmartinez287:
## The Issue
Prototypes a pricing page for DDEV

Encourages employees to talk about buying a license code maybe a UUID string we can keep track of internally.

It's a rough prototype but something that might be worth exploring.

### DIFF
--- a/src/components/Header.astro
+++ b/src/components/Header.astro
@@ -91,17 +91,24 @@ import SponsorsBanner from "./SponsorsBanner.astro"
           >
         </li>
         <li class="mx-4">
+          <a
+            class="font-bold whitespace-nowrap dark:text-white"
+            href="/pricing"
+            >Pricing</a
+          >
+        </li>
+        <li class="mx-4">
           <a class="font-bold whitespace-nowrap dark:text-white" href="/blog" rel="prefetch"
             >Blog</a
           >
         </li>
-        <li class="mx-4">
+        <!-- <li class="mx-4">
           <a
             class="font-bold whitespace-nowrap dark:text-white"
             href={GITHUB_URL}
             target="_blank">GitHub</a
           >
-        </li>
+        </li> -->
         <li class="mx-4">
           <a
             class="font-bold whitespace-nowrap dark:text-white"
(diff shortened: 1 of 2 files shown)


bmartinez287:
https://58005808.ddev-com-front-end.pages.dev/pricing/

rfay:
This is really nice, and it's a good idea. 

It pushes us to be more explicit about partnership levels, and that's fantastic.

Right now, there are probably too many options here and it can be simplified. 

And we need to feel the push to be more explicit and sort that out!

I'd love if if you'd take a look at a thermometer like thedroptimes.com now has using our constantly updated data from https://github.com/ddev/sponsorship-data/blob/main/data/all-sponsorships.json - We need to display that on the website, preferably on the github README and also somehow in the remote-config (https://github.com/ddev/remote-config)

stasadev:
That looks great!

I think it's important to include a free option in the list, as leaving it out might create the false impression that DDEV is a paid app.

bmartinez287:
The challenge with displaying a free option is that some employers might default to that if that's an option when looking at that page. Like, if I requested DDEV as a tool for work they likely suggest I use the free one.
That said, I would like to avoid hurting new user adoption but unsure of how to word it so it seems to be a greatly reduced list of benefits.

stasadev:
I agree with you from this point of view.

bmartinez287:
Maybe we could create a limitation on the amount of projects DDEV can start.
If it detects that DDEV list has more than 2 or 3 projects at any given time and a user is on the free tier it will block you from creating more until you add a license key. 

The license key could be something that's added to the .env file.

The paid tier gives you unlimited projects, which is easy to fall into if you are using ddev extensively and not a blocker for new users. This would encourage heavy users to switch to a paid tier and for small use cases they could use it for free.

rfay:
I'm interested in *adding* things for paid users (but may be too complex to implement) but not in taking anything away from anybody.

bmartinez287:
Makes sense, how about making the ddev for JavaScript or the feature after that premium?

GuySartorelli:
Looks very nice visually, as a proof of concept this is great.

I agree with the concern that not including a free tier (or even just a separate paragraph) makes it look like DDEV is not free to use at all anymore.

Even just a paragraph below the teir list saying something like "DDEV is still free to use, but some features will only be available under a paid plan" would resolve that for me. You could also include there a call to action to support DDEV development by buying a license. "Buying a license gives you extra features, and helps to support the ongoing maintenance and development of DDEV"

rfay:
Thanks for this initiative, it's important and helps shape the conversation.

I haven't yet figured out how this would work technically or policy-wise. If anybody has suggestions about either of those they're welcome! It's one thing to put up a page, another entirely to implement the things it calls for.

stasadev:
Closed in favor of #655, thanks!

PR #337, opened 2025-03-03
Added Affinity Bridge logo to Sponsors page.

marklabrecque-ab:
As discussed, this simply adds the Affinity Bridge logo to the sponsors page.

Thanks for raising this issue, Randy!


rfay:
I don't understand why, but I don't see this showing up on my local although the browser shows it being downloaded.

image

bmartinez287:
that's likely because it's missing from this page. https://github.com/ddev/ddev.com/blob/main/src/components/FeaturedSponsors.astro

bmartinez287 (approved):

bmartinez287:
[public/logos/affinity-bridge-dark.svg](https://github.com/ddev/ddev.com/pull/337/files#diff-ca801164ab7bc6aba683dc29be9887131368f81fbf4997e590ff1512c8d6d5a9) says dark because the colors are dark but that's not why we name them dark that's the opposite of it.

Dark mode files have a white outline so they display on dark screens.

rfay:
Am I blind that I can't see it on my local, or doing something wrong?

stasadev (approved):
Should be good now:

rfay:
Thanks so much for making this easy @marklabrecque-ab !

PR #360, opened 2025-04-16
AMD64 on ARM: Add better example for the run-webserver case

rfay:
## The Issue

The new blog on running AMD64 on ARM didn't have a good example for running ddev-webserver

Add example.

Rendered at https://20250416-amd64-touchup.ddev-com-front-end.pages.dev/blog/amd64-with-rosetta-on-macos/

### DIFF
--- a/src/content/blog/amd64-with-rosetta-on-macos.md
+++ b/src/content/blog/amd64-with-rosetta-on-macos.md
@@ -74,7 +74,7 @@ Now on `ddev restart` you'll be running an AMD64 web container and `ddev npm ins
 
 Test it with `ddev exec arch`, you'll now get `x86_64`.
 
-(If you have trouble, so a `ddev debug rebuild` to make it forget the old version of the Dockerfile build.)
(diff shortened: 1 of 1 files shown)


src/content/blog/amd64-with-rosetta-on-macos.md
```diff
@@ -59,9 +70,15 @@ services:
     platform: linux/amd64
 ```
 
+Now on `ddev restart` you'll be running an AMD64 web container and `ddev npm install gifsicle` will work just fine. And your colleagues who are on Intel processors will have no trouble with this configuration.
+
+Test it with `ddev exec arch`, you'll now get `x86_64`.
+
+(If you have trouble, so a `ddev debug rebuild` to make it forget the old version of the Dockerfile build.)
```
stasadev, line 77:
```suggestion
(If you run into issues, try `ddev debug rebuild` to clear the old Dockerfile build cache.)
```

stasadev (approved):
Thank you!

PR #363, opened 2025-04-29
feat: add sponsors banner to ddev.com, fixes #339

bmartinez287:
## The Issue
Adds Sponsor banner on ddev.com
## How This PR Solves The Issue
Solves #339 
Screenshot 2025-04-29 at 11 19 54 AM
Screenshot 2025-04-29 at 11 20 08 AM

PR #372, opened 2025-05-08
refactor: migrate sponsors banner to component with caching, for #363

stasadev:
## The Issue

- #363

The sponsors banner has a few problems:

Sponsorship data isn't cached, and there’s no fallback if GitHub is unavailable:

image

Pressing F5 or moving from page to page causes the whole site to jump:

jump

1. Adds initial cache for all sponsorships during `npm run build` (so GitHub failure doesn't break it) and then caches fresh data for a day.

2. I tried several ways to make the banner appear immediately. Nothing worked, I asked ChatGPT to refactor it into an Astro component, which stopped the jumping. This was annoying.

1. Open https://20250508-stasadev-sponsors-b.ddev-com-front-end.pages.dev/
2. Check that it doesn't jump
3. Open Dev panel (F12)
4. Check network tab for Fetch/XHR requests, it should fetch from GitHub only the first time
5. Check application tab for `ddevSponsorshipData` in local storage
6. Delete `ddevSponsorshipData`, open network tab again, and check for new fetch from GitHub

Note: I deleted `src/scripts/sponsors-banner.js` as not needed.


rfay:
Thanks!

PR #379, opened 2025-06-10
fix: update sponsorship cache, fixes #378

### LINKED ISSUE #378: Monthly DDEV Sponsorship income never updates in a given browser
Our current reported sponsorship income (in a browser which has visited ddev.com) is never updated.

Image

Current from https://raw.githubusercontent.com/ddev/sponsorship-data/refs/heads/main/data/all-sponsorships.json is $7069

But on my browser I see $7569. 

(There also seems to be a discrepancy in that our calculated $7069 is down from previous calculations in sponsorship-data, but I'll chase that elsewhere.)

stasadev:
## The Issue

- #378

The logic was missing one `else` condition, where cached data should be updated.

- Refactors to several functions to make more readable.
- Uses non blocking async/await.
- Sets the cache for 12 hours.

https://20250610-stasadev-sponsorshi.ddev-com-front-end.pages.dev/

### DIFF
--- a/src/components/SponsorsBanner.astro
+++ b/src/components/SponsorsBanner.astro
@@ -140,8 +140,8 @@ function formatCurrency(value: number): string {
 <script is:inline>
   const goal = 12000;
   const cacheKey = 'ddevSponsorshipData';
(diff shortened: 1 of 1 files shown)


rfay:
Let's give it a go!

PR #399, opened 2025-07-21
Update foundation.astro with current info

rfay:
## The Issue

@stasadev noted that there was incorrect information about Randy not receiving funds

Thanks!

Review updated foundation.astro at https://20250712-rfay-foundation.ddev-com-front-end.pages.dev/foundation/


stasadev (approved):
Thank you, looks good to me!

PR #404, opened 2025-07-23
docs: add command to update wsl for Windows

adiati98:

When installing DDEV, I got an error message saying that DDEV can't be installed. After [a discussion and troubleshooting on Discord](https://discord.com/channels/664580571770388500/1397288639503007825), re-installing the WSL with `wsl --install` and updating with `wsl --update` solved the problem. DDEV was installed smoothly.

Based on this, in some cases, running `wsl --update` command after installing WSL on Windows might be required to install DDEV successfully. 

This PR adds the command to the [Get Started](https://ddev.com/get-started/) page.

### DIFF
--- a/src/components/quickstart/Windows.astro
+++ b/src/components/quickstart/Windows.astro
@@ -32,13 +32,11 @@ const { latestVersion } = Astro.props
 
   <p>You’ll probably need to reboot.</p>
 
-  <p>After reboot, update your WSL:</p>
+  <p>If you had previously installed WSL, update it:</p>
 
   <Terminal type="powershell" code={`> wsl --update`} />
-
-  <p>Keeping WSL updated helps to ensure that DDEV installs successfully.</p>
-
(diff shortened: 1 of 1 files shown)


src/components/quickstart/Windows.astro
```diff
@@ -31,7 +31,13 @@ const { latestVersion } = Astro.props
 
   <p>You’ll probably need to reboot.</p>
 
-  <p>After reboot, install an Ubuntu distro:
+  <p>After reboot, update your WSL:</p>
```
rfay, line 34: I'm conflicted on this, because of course if you just installed and rebooted, you don't need to update. OTOH we're trying to keep this as simple and straightforward as possible.

adiati98: Then I think we don't need to add this, assuming that the info on this page is for folks who are completely new to WSL. 🙂

PR #407, opened 2025-07-23
Add Passbolt as a featured sponsor

LeonardoPagani:


stasadev (approved):
Thank you very much @LeonardoPagani !

Preview: https://16931453.ddev-com-front-end.pages.dev/

PR #418, opened 2025-08-22
refactor: make sponsor banner more compact

stasadev:
## The Issue

The banner takes too much height:

image

Adjusts the layout for different widths.

https://20250822-stasadev-sponsor-ba.ddev-com-front-end.pages.dev/

Open developer tools (F12), enable responsive mode, and resize the viewport to confirm banner height adapts correctly.

banner


rfay:
I like it!

PR #428, opened 2025-09-19
docs: Give better instructions for WSL2 users to work inside distro

rfay:
## The Issue

We occasionally see Windows users stumble and create their project on Windows, which is really slow. This is called out carefully in docs.ddev.com but not carefully called out in get-started here.

Make clearer that they need to work in the distro

Review rendered Windows instructions at https://20250919-rfay-windows-get-st.ddev-com-front-end.pages.dev/get-started/

### DIFF
--- a/src/components/quickstart/Windows.astro
+++ b/src/components/quickstart/Windows.astro
@@ -85,7 +85,11 @@ const { latestVersion } = Astro.props
 
   <p>Tip: Watch the <a href="/blog/watch-ddev-local-from-scratch-with-windows-wsl2/">video tutorial</a> showing all of this.</p>
 
-  <p>Inside the WSL2 "DDEV" distro create a directory for your project or clone your project. (<strong>This is not on the C:\ drive; Use the "DDEV" or "Windows Terminal" apps to access it. Or use <code>Explorer > Linux > DDEV > /home/&lt;you&gt;</code> to see the files.</strong>) </p>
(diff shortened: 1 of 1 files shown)


src/components/quickstart/Windows.astro
```diff
@@ -73,7 +77,105 @@ const { latestVersion } = Astro.props
...
 </div>
 
-<Examples itemNumber="3/3" terminalType="powershell-ubuntu" terminalSymbol="$" />
+<div class="prose dark:prose-invert">
+  <h2 class="text-3xl mt-24">
+    <small class="block font-mono text-xs text-gray-500">{itemNumber}</small>Create a
+    Project
+  </h2>
+
+  <p>Inside the WSL2 "DDEV" distro (<strong>not on Windows</strong>), create a directory for your project or clone your project:</p>
```
stasadev, line 86: "not on Windows" - if I was Windows only user, I would not understand this.

Maybe something like:

not in `C:\` or `D:\` drive, use Explorer > `Linux` in the sidebar > `DDEV` > `/home/<you>` to see the actual files.

stasadev (commented):
What about adding a tip about https://ddev.com/blog/watch-ddev-local-from-scratch-with-windows-wsl2/ here in "Get Started" for Windows?

stasadev (approved):
Thank you, this is the best thing we can do for Windows users.

PR #431, opened 2025-09-19
blog: Update the older hostname resolution blog with training video

rfay:
We did a contributor training on the hostnames/DNS/ddev.site topic this week, but it mostly covers what's in the existing article, so updating the article with the video.

This is also the first test with a forked PR.

Here's the rendered URL. https://pr-431.ddev-com-fork-previews.pages.dev/blog/ddev-name-resolution-wildcards/


stasadev (approved):
> Here's the rendered URL, but it may change with additional commits, since the new forked pr approach doesn't seem to have a stable URL

Isn't this a stable URL?

https://pr-431.ddev-com-fork-previews.pages.dev/

PR #450, opened 2025-10-15
fix(codespaces): use lightweight debian-12 image

stasadev:
## The Issue

Implements the same thing as in:

- https://github.com/ddev/ddev/pull/7713


stasadev:
No, approval for ddev covers this too.

PR #487, opened 2025-12-13
feat: copy button on blog pages and getting started pages

bmartinez287:
## The Issue

Adds a copy button for code.

Use Gemini to determine how to integrate it with the getting-started page.

## Manual Testing

Review at, for example:
* https://20251213-bmartinez-copybutto.ddev-com-front-end.pages.dev/blog/ddev-on-linux-in-10-minutes/
* https://20251213-bmartinez-copybutto.ddev-com-front-end.pages.dev/get-started/

Screenshot 2025-12-13 at 4 34 53 PM

### DIFF
--- a/src/styles/global.css
+++ b/src/styles/global.css
@@ -178,7 +178,7 @@ pre.astro-code > button.copy {
   position: absolute;
   top: 0.65rem;
   right: 0.85rem;
-  padding: 0.3rem 0.6rem;
+  padding: 0.25rem 0.7rem 0.2rem;
   border: none;
   border-radius: 4px;
   cursor: pointer;


bmartinez287:
1.  It was a typo, so that one is easy.
2. This one is tricky, but I improved it a bit by making the padding even on each side.
3. Not sure if any copy packages do that. It would be cool, but it would likely be challenging since there are so many things specific to each language.

mattstein:
Thanks @bmartinez287! I realize the copy area is an annoying thing to deal with, it just seems like a minor barrier to someone getting started.

Do you think it’s possible to match the padding on blog post pages? That looks more optically correct to me ([example](https://4065a3e7.ddev-com-front-end.pages.dev/blog/tailscale-router-ddev-addon/)):



That left padding would ideally match the right side (1), with a little bit of padding at the bottom of the window so a one-line example still gets even spacing to the right and bottom of the copy button (2). (You could probably scoot the copy button up slightly and aim to align the baseline of the code and copy text.) Nitpicky again but it’d feel more balanced and better match the other less-fancy code blocks.

rfay:
Thank you for making this happen @bmartinez287 ! Thanks for still paying attention @mattstein !

PR #557, opened 2026-02-24
feat: replace blog pagination with "Load More" button and optional auto-scroll, improve noscript support

stasadev:
## The Issue

Blog post listings used prev/next pagination links requiring full page navigations to browse content. Search results were capped at 10 with no way to see more, and did not restore when navigating back. Several site features (terminal animation, platform picker, code copy buttons) had no meaningful fallback without JavaScript.

**"Load More" button and optional auto-scroll for blog listings**

Replaces prev/next pagination with a fetch-HTML-fragment approach. Each listing page (index, category, author) renders its first 15 posts normally. A `data-next-url` attribute on the grid container points to the next pre-built static page. Clicking "Load More" fetches that URL, extracts the post cards via `DOMParser`, and appends them to the current grid. The next `data-next-url` is then read from the fetched page to continue the chain.

An "Auto-load on scroll" checkbox lets users switch to infinite scroll mode. The preference is saved to `localStorage` and applied on every blog listing page. It defaults to "Load More" button mode, keeping the footer accessible. The toggle is visible on all listing pages (including short ones) so the preference can always be changed.

This keeps every page's initial payload at 15 posts regardless of total post count, preserves full Astro image optimization on all cards (no client-side re-rendering), and leaves all paginated static routes intact as valid URLs and noscript fallbacks.

All logic is centralised in a new `BlogPostGrid` component. `Paging.astro` is removed as it is no longer referenced anywhere.

**Progressive search results**

Search results now load in batches of 10 as the user scrolls, with all matches held in memory after each query. Previously results were hard-capped at 10. The input value is also checked on page initialisation so results are restored correctly after browser back navigation.

**Noscript fallbacks**

- `BlogPostGrid`: prev/next pagination links are rendered inside `<noscript>` tags; the "Load More" button is hidden by default and only revealed by JS
- `AnimatedTerminal`: a static `ddev describe` output is shown via a second `<pre>` element; toggled by `html[data-theme]` (set by an inline head script before body renders) so there is no flash when JS is enabled
- `get-started.astro` PlatformPicker: CSS `:has()` rules make radio-based platform selection work without JS
- Code copy buttons: hidden via `html:not([data-theme])` selector when JS is absent

**Table wrapping**

A new rehype plugin (`rehype-wrap-tables.mjs`) wraps markdown tables in a `div.table-wrapper` for horizontal scroll on mobile.

- Visit https://pr-557.ddev-com-fork-previews.pages.dev/blog/ and click "Load More" to verify posts append without page navigation
- Check the "Auto-load on scroll" checkbox and scroll down to verify posts load automatically
- Reload the page and confirm the checkbox is still checked and scroll mode is still active
- Visit https://pr-557.ddev-com-fork-previews.pages.dev/blog/category/guides/ and https://pr-557.ddev-com-fork-previews.pages.dev/blog/author/randy-fay/ and click "Load More"
- Check Network tab: each fetch returns 200, not 301
- Disable JS (https://developer.chrome.com/docs/devtools/javascript/disable) and confirm:
  - Blog listing shows prev/next pagination links
  - https://pr-557.ddev-com-fork-previews.pages.dev/get-started/ platform picker still works via click
  - Terminal on the home page shows static `ddev describe` output with no flash on re-enable
  - Code copy buttons are not visible
- Search https://pr-557.ddev-com-fork-previews.pages.dev/search/ for a term with many results (e.g. "windows") and scroll to load beyond the first 10
- Navigate to a result, press back, confirm results are restored

## Automated Testing Overview

No automated tests added; all changes are client-side JS and build-time HTML transforms.

## Release/Deployment Notes

All paginated static routes (`/blog/2/`, `/blog/category/x/2/`, etc.) remain valid URLs. No redirects or deployment config changes needed.

### DIFF
--- a/src/components/BlogPostGrid.astro
+++ b/src/components/BlogPostGrid.astro
@@ -1,8 +1,9 @@
 ---
 /**
- * Renders a grid of blog post cards with a "Load More" button for fetching
- * additional pages via HTML fragment fetching.
- * Pass `nextUrl` to enable loading the next page on button click.
+ * Renders a grid of blog post cards with a "Load More" button or infinite scroll
+ * for fetching additional pages via HTML fragment fetching.
+ * Pass `nextUrl` to enable loading more posts.
+ * The user's preference (button vs auto-scroll) is persisted in localStorage.
  */
 import type { CollectionEntry } from "astro:content"
 import BlogPostCard from "./BlogPostCard.astro"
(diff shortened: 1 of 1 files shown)


rfay (commented):
That's amazing!

stasadev:
> The only "issue "with infinity scroll is you lose access to the page footer as you have to scroll increasing further to reach it. With this is mind, are there any links that would be better in the header?

It might be better to use a "Load More" button that fetches and appends items instead of loading them automatically.

PR #599, opened 2026-04-03
build: split preview forks into two workflows, don't use `pull_request_target`

stasadev:
## The Issue

- #598

https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

Running build with `pull_request` won't trigger any workflows for first-time contributors.

`pull_request_target` is still used for comments, where it's safe because it doesn't use checkout.


rfay (approved):
Thanks for doing this. Claude says it's great. Preview looks fine.  Preview link gets generated and works.

PR #607, opened 2026-04-08
fix: normalize hyphenated search terms, fixes #602

### LINKED ISSUE #602: Search needs tweak to prefer keywords in title or other frontmatter
Right now if you search for `add-on` or `add-ons` at https://ddev.com/search/ the key [add-on maintenance guide](https://ddev.com/blog/ddev-add-on-maintenance-guide/) is 12th in the list. It should be first :) 

If you search for `addon`, the maintenance guide is second, but should be first.

stasadev:
## The Issue

Searching for "add-on" did not match content containing "addon" (and vice versa), because MiniSearch treated them as distinct tokens.

- Fixes #602

Configure MiniSearch `tokenize` and `processTerm` options in both the index builder (`src/lib/search-index.js`) and the search page (`src/pages/search.astro`) to keep hyphenated words as single tokens and strip hyphens before indexing/querying, so "add-on" and "addon" resolve to the same term.

Run search https://pr-607.ddev-com-fork-previews.pages.dev/search/

### DIFF
--- a/src/lib/search-index.js
+++ b/src/lib/search-index.js
@@ -118,7 +118,13 @@ export default function searchIndex(config) {
 
         // Generate the index in the format we need
         let index = await getSearchIndex(items, {
+          fields: ["title", "heading", "text"],
           storeFields: ["title", "heading", "pubDate", "modifiedDate"],
+          // Keep hyphenated words as single tokens, then strip the hyphen so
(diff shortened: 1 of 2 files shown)


rfay (approved):
Sure solves #602 , thanks!

PR #640, opened 2026-05-16
feat(sponsor): add /sponsor page with live progress + source breakdown

js360000:
## Summary

Adds a new `/sponsor` page focused entirely on financial sponsorship, addressing the requirements in #631 directly:

- ✅ CTAs for one-time and monthly recurring donations, with monthly recurring as the emphasis.
- ✅ Sub-CTAs for GitHub Sponsors, PayPal, and direct invoice — each with copy targeted at its likely audience.
- ✅ Live data pulled from `https://ddev.github.io/sponsorship-data/data/all-sponsorships.json` at build time, with a typed fallback snapshot so the build never breaks on a transient fetch failure.
- ✅ Source breakdown table — monthly recurring equivalent across GitHub Sponsors, monthly invoiced, annual invoiced (normalized to monthly), and PayPal.
- ✅ Featured organizations block (reusing the existing `FeaturedSponsors` component, no new content infrastructure needed).
- ✅ The "bubble" sponsor grid from `/support-ddev/` moved to the bottom (as suggested in the issue).
- ✅ 501(c)(3) note linking out to `/foundation`.

## Live progress block

The hero block reads, with current data:

> **$9,456 / $12,000** — 78.8% of the $12,000/month goal  
> + $1,603 vs. one week ago  
> Data updated daily · last update May 16, 2026 · [view raw JSON]

The progress bar uses the existing emerald → cyan gradient palette so it sits inside the site's visual language.

## /support-ddev/ — unchanged in scope, points to /sponsor for finance

`/support-ddev/` opens with a banner directing financial sponsors to `/sponsor`, and its in-page table of contents entry for "Financial Support" links to `/sponsor` directly. The existing financial section on `/support-ddev/` stays in place so no external links break — `/sponsor` is the canonical entry going forward, but old bookmarks still work.

## Daily-data freshness

The data fetch happens at build time, so the page is always accurate as of the last deploy. Cloudflare Pages can be configured with a daily build hook (or webhook triggered by the `sponsorship-data` repo on its daily snapshot commit) to pull the latest snapshot. If real-time freshness becomes important between deploys, a small client-side refresh script can be added as a follow-up — happy to do that in a follow-up PR if useful.

## Code notes

- Direct, factual copy throughout per `AGENTS.md` — no marketing superlatives.
- Dark-mode aware, follows the existing `dark:` utility conventions.
- `astro check` passes cleanly for both files; full `astro build` completes successfully and produces `dist/sponsor/index.html` with all live data baked in.
[...]

Issue #631, opened 2026-05-08
New ddev.com/sponsor page


ultimike:
DDEV.com needs a main "Sponsor" page that we can send potential financial sponsors to as a starting point. 

Currently, https://ddev.com/support-ddev/ acts as the main entry point for sponsors, but then veers off to either:

1. https://github.com/sponsors/ddev
2. https://www.paypal.com/donate/?hosted_button_id=MCNCSZHC7LHSQ
3. https://ddev.com/contact (for sponsors who prefer traditional invoices)

I think we need a cleaner, simpler, more engaging landing page for potential sponsors. I'm not sure if it would be best to keep https://ddev.com/support-ddev/ as it covers more than just financial support. Perhaps we can move some of the financial bits from that page and move them to the new /sponsor page.

IMHO, the new /sponsor page should include:

1. CTAs for one-time donations and monthly recurring donations, with an emphasis on the latter. 
2. For monthly recurring donations, sub-CTAs for the donator's method of choice (GitHub, PayPal, or invoicing.) 
3. Data from https://github.com/ddev/sponsorship-data, updated daily. This should include a breakdown of monthly recurring donations by source (GitHub, PayPal, invoicing.) 
4. A top 10 list of our largest monthly recurring sponsors (including clickable logos) with some text to incentive larger organizations to get on the top 10 list. Another option is to have the top 5 sponsors by continent.
5. I really like the "bubble" list of sponsors on https://ddev.com/support-ddev/ - maybe we can move that to the bottom of the new /sponsor page.
6. The sub-CTA for invoicing should either lead directly to an email address or a webform. 

I'm not crazy about the fact that https://github.com/sponsors/ddev lists the same $12000/month goal but only includes GitHub sponsors as progress. 

@rfay informed me that @bmartinez287 created [a mockup](https://20250219-bmartinez-ddev-pric.ddev-com-front-end.pages.dev/pricing/) a while ago showing a more direct landing page. I like the design, but I want to make it clear that any amount is welcome. 

I'd love some more feedback some ideas for what can/should be included on the new landing page before I work with Claude Design (I've been looking for an excuse to play with it!) on a mockup or two.

thanks,
-mike

rfay:
Actually most of our things point to https://github.com/sponsors/ddev as the primary sponsors landing, and it's not great.

I wonder if a page on ddev.com could use the github API to sign people up, so we'd have a single consolidated place.

deviantintegral:
[...]

PR #644, opened 2026-05-18
Prototype for ddev sponsorship page v3.

bmartinez287:
This Page includes another demo on how a prototype page for issue #631 could look.
Very similar in some ways to #640 but with a different spin. The community and the ddev team could cherry pick elements from either or add comments on which approach might lead to the MVP.

/sponsors is the url for this page.

go to /sponsors is the url for this page and provide some feedback.

## Related Issue Link(s)
#640 #329 #631

### DIFF
--- a/src/pages/sponsors.astro
+++ b/src/pages/sponsors.astro
@@ -1,0 +1,360 @@
+---
+import Layout from "../layouts/Layout.astro"
+import Heading from "../components/Heading.astro"
+import Sponsors from "../components/Sponsors.astro"
+
+const title = "Sponsor DDEV"
+
+const individualTiers = [
+  {
+    price: "$50",
+    name: "Supporting Sponsor",
+    description:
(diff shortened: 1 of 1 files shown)


bmartinez287:
This page can be previewed here https://b87885d9.ddev-com-front-end.pages.dev/sponsors/

tyler36:
Great work @bmartinez287 

This is the classic layout, but for me, I find there are way too many options, especially under the "Organizational Level" heading. As a developer, I prefer #640 (highlights "peer approval" and I love a nice chart ) but I think this PR might be an easier sell to managers.

DDEV has a lot of users in EU and I wonder if current translation software is able to convey the difference in support levels displayed on the cards.

Also noting that #640 uses `/sponsors` and this PR uses `/sponsor`.

rfay:
Closed in favor of #655, thanks!

PR #651, opened 2026-06-15
chore: remove Optasy from featured sponsors

rfay:
Optasy has stopped supporting DDEV.

Remove from featured sponsors.

Review at https://pr-651.ddev-com-fork-previews.pages.dev/

I left the svg files there in case they come back.


stasadev (approved):
Looks good to me.

> I left the svg files there in case they come back.

Agreed.

Issue #647, opened 2026-05-25
Update DDEV Foundation sponsorship levels


ultimike:
Working with the [existing sponsorship levels](https://github.com/sponsors/ddev) and some ideas from myself and others (including help from Claude), I'd like to propose this updated list of sponsorship levels and benefits (in the form of a "sponsorship prospectus".)

This sponsorship prospectus can be used as a PDF sent directly to potential contributors. The updated sponsorship levels will be one of the main focuses of the soon-to-be-updated sponsor home page (#631). 

This task is a prerequisite for #631 

[Updated Sponsorship Prospectus here](https://github.com/ddev/sponsorship-data/blob/68567dc3cb9a619aeca5dfd0d2ad024d7c3f1d9c/docs/sponsorship-levels.md)

-------------------------------------------
~# DDEV Foundation Sponsorship Prospectus~

~## What is DDEV?~

~DDEV is the open-source local development environment trusted by over 19,000 active developers worldwide. It makes local web development reliable, fast, and consistent for PHP, JavaScript, and HTML developers — whether they're working with Drupal, WordPress, Laravel, TYPO3, Craft CMS, Magento, or any other web platform.~

~Since its origins in 2016, DDEV has grown into a vital piece of infrastructure for thousands of development teams, agencies, and individual developers. With 3,600+ GitHub stars, 222 releases, 399 contributors, and a vibrant community on Discord, DDEV is one of the most actively maintained and widely adopted local development tools in the open-source ecosystem.~

~## The DDEV Foundation~

~The DDEV Foundation is a US 501(c)(3) nonprofit corporation registered in Colorado. It serves as the organizational home for the DDEV project — owning the infrastructure, managing funds, and ensuring the project's long-term sustainability through community governance and a Board of Directors.~

~The Foundation's funds go directly toward infrastructure costs and paying DDEV's maintainers. There are no shareholders, no venture capital expectations, and no hidden agendas. Every dollar of sponsorship goes toward making DDEV better for the developers who depend on it.~

~## Why Recurring Monthly Sponsorship Matters~

~Open-source projects like DDEV face a fundamental challenge: the software is free, but building and maintaining it is not. DDEV's two full-time maintainers — Randy Fay and Stas Zhuk — dedicate their professional lives to keeping DDEV reliable, secure, and constantly improving. That work requires sustainable, predictable funding.~

[...]

PR #655, opened 2026-06-23
feat(sponsor): replace support-ddev page with sponsor page

bmartinez287:
Uses the text and findings on 

https://github.com/ddev/ddev.com/issues/647
https://github.com/ddev/sponsorship-data/blob/68567dc3cb9a619aeca5dfd0d2ad024d7c3f1d9c/docs/sponsorship-levels.md?plain=1

Review at https://20260622-bmartinez-ddev-supp.ddev-com-front-end.pages.dev/support-ddev/

### DIFF
--- a/src/pages/contact.astro
+++ b/src/pages/contact.astro
@@ -80,6 +80,25 @@ const title = `Contact Us`
           </div>
         </section>
 
+        <section id="sponsorship" class="prose dark:prose-invert mb-12 scroll-mt-24">
+          <h2>Sponsorship &amp; invoicing</h2>
+          <p>
+            Prefer to pay by invoice or bank transfer instead of GitHub Sponsors
+            or PayPal? Email us at <a href={EMAIL_URL}>support@ddev.com</a> and include:
+          </p>
+          <ul>
+            <li>Organization / legal name and billing address</li>
+            <li>Billing contact email</li>
+            <li>Monthly amount or sponsorship tier</li>
+            <li>Preferred payment method (ACH, Wise, bank transfer, IBAN, SWIFT, or check)</li>
+            <li>Any purchase order, VAT ID, or tax forms (for example a W-9) you need</li>
+          </ul>
+          <p>
+            We'll reply with an invoice and payment details. As a US 501(c)(3)
+            nonprofit, sponsorships to the DDEV Foundation may be tax-deductible.
+          </p>
+        </section>
+
         <section class="prose dark:prose-invert">
           <h2>DDEV Foundation Information</h2>
           <p>
--- a/src/pages/sponsor.astro
+++ b/src/pages/sponsor.astro
@@ -1,6 +1,5 @@
 ---
 import Layout from "../layouts/Layout.astro"
-import Heading from "../components/Heading.astro"
 
 const title = `Sponsor DDEV`
 
@@ -51,7 +50,7 @@ const sponsorshipTiers = [
     description: "Great for large organizations.",
     rewards: [
       "All the benefits of lower tiers",
-      'Your logo and link on [ddev.com/support-ddev](https://ddev.com/support-ddev/) and [github.com/sponsors/ddev](https://github.com/sponsors/ddev) in the "Featured Sponsors" section',
+      'Your logo and link on [ddev.com/sponsor](https://ddev.com/sponsor/) and [github.com/sponsors/ddev](https://github.com/sponsors/ddev) in the "Featured Sponsors" section',
       "Prioritized bug reports",
       "DDEV Tip of the Day names and thanks you — seen by thousands of developers every day in the terminal",
       "Sponsor spotlight blog post on ddev.com",
--- a/src/pages/sponsor.astro
+++ b/src/pages/sponsor.astro
@@ -6,6 +6,16 @@ const title = `Sponsor DDEV`
 const linkClass =
   "font-bold text-blue-700 underline hover:text-blue-900 dark:text-blue-400 dark:hover:text-blue-300"
 
+const paypalCta = {
+  label: "Sponsor via PayPal",
+  href: "https://www.paypal.com/donate/?hosted_button_id=MCNCSZHC7LHSQ",
+}
+
+const invoiceCta = {
+  label: "Contact us to sponsor",
+  href: "/contact#sponsorship",
+}
+
 const parseReward = (text: string) => {
   return text.split(/(\[[^\]]+\]\([^)]+\))/g).map((part) => {
     const match = part.match(/^\[([^\]]+)\]\(([^)]+)\)$/)
@@ -34,6 +44,7 @@ const sponsorshipTiers = [
       "A heartfelt thank-you from the DDEV team",
       "Sponsor badge on your GitHub profile",
     ],
+    secondaryCta: paypalCta,
   },
   {
     amount: "100",
@@ -42,10 +53,11 @@ const sponsorshipTiers = [
     featured: true,
     rewards: [
       "All the benefits of lower tiers.",
-      "Your logo and link in the DDEV README and on ddev.com",
+      "Your logo and link in the DDEV [README](https://github.com/ddev/ddev#featured-sponsors) and on [ddev.com](/#supporters) in the \"Featured Sponsors\" section",
       "Your entire team gets to use [coder.ddev.com](https://coder.ddev.com/) to experience DDEV in the cloud. See [description](/blog/coder-ddev-com-announcement/).",
       "Free [Diffy](https://diffy.website/) visual regression testing Pro plan license for one year ($800 value).",
     ],
+    secondaryCta: paypalCta,
   },
   {
     amount: "500",
@@ -53,11 +65,12 @@ const sponsorshipTiers = [
     description: "Great for large organizations.",
     rewards: [
       "All the benefits of lower tiers",
-      'Your logo and link on [ddev.com/sponsor](https://ddev.com/sponsor/) and [github.com/sponsors/ddev](https://github.com/sponsors/ddev) in the "Featured Sponsors" section',
+      'Your logo and link on [ddev.com/sponsor](/sponsor/) and [github.com/sponsors/ddev](https://github.com/sponsors/ddev) in the "Featured Sponsors" section',
       "Prioritized bug reports",
       "DDEV Tip of the Day names and thanks you — seen by thousands of developers every day in the terminal",
-      "Sponsor spotlight blog post on ddev.com",
+      "Sponsor spotlight blog post on [ddev.com/blog](/blog/)",
     ],
+    secondaryCta: invoiceCta,
   },
   {
     amount: "1,000",
@@ -68,6 +81,7 @@ const sponsorshipTiers = [
       "We join your company chat app (Slack, Teams, etc.) for direct help and support",
       "Direct line to the maintainers for your team's DDEV questions",
     ],
+    secondaryCta: invoiceCta,
   },
 ]
 ---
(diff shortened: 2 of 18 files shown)


src/pages/support-ddev.astro
```diff
@@ -1,130 +1,198 @@
...
+      "A heartfelt thank-you from the DDEV team",
+      "Sponsor badge on your GitHub profile",
+      "Universities and educational institutions: a one-time annual contribution of $400 qualifies for this tier",
+    ],
+  },
+  {
+    price: "$100",
+    name: "Featured Sponsor",
+    description: "Great for small agencies.",
+    rewards: ["Your logo and link in the DDEV README"],
```
stasadev: We'll need to reorganize our SVGs in this repo for this to have split logic.

stasadev (commented):
That's great!

My nitpicks:

1. When I open the page, I need to scroll before I can see the sponsor tiers. I wonder if it would be better to show a short introductory paragraph first, then the tiers, and move the detailed information below.
2. The $500 and $1000 tiers are very unlikely to use GitHub Sponsors or PayPal. It would be better to add a direct billing option for those tiers.
3. We need to show both USD and EUR, just to make it clear that people are not restricted to USD.

src/pages/support-ddev.astro
bmartinez287: oooo probably, seems like a good idea even though that is one extra task to take into consideration.

rfay:
Thanks for pushing this forward!!! So great.

* "DDEV Foundation Sponsorship Prospectus" is not the name we want for this page. Better "Sponsor DDEV!"
* I think there's too much text before the offerings. It should be really minimal. Focus on what they get in the offerings, rather than how important it is to DDEV.
* The extra text/explanations can go below the offerings.
* We don't normally propose PayPal as a primary funding source, and especially for donors over $25
* We don't currently do Venmo, but if we added that it could be added to the page.

ultimike:
I echo both @stasadev's and @rfay's comments above. Two additional thoughts:

1. I'd move the first two paragraphs to below the sponsorship levels.
2. [This is the most recent version of the sponsorship levels](https://github.com/ddev/sponsorship-data/blob/main/docs/sponsorship-levels.md) - it includes Diffy and some other stuff that the link above don't have. Sorry for the confusion.

Thanks, @bmartinez287!!!

-mike

bmartinez287:
we do offer venmo. The paypal link allows you to use a credit card, paypal or venmo to pay.
Screenshot 2026-06-23 at 10 16 43 PM

stasadev:
> we do offer venmo.

https://support.trolley.com/s/article/Venmo-FAQ

> 1. Which countries is Venmo available in?
> Currently, Venmo is only available in the United States, Puerto Rico, Guam and the Virgin Islands.

I don't even see that option, and the EU doesn't see it.

---

I would simply rename the button to "PayPal / Card"

src/pages/support-ddev.astro
```diff
@@ -1,130 +1,245 @@
...
+              your tier. For example, a $1,200/€1,200 annual contribution qualifies
+              for the $100/€100 per month Featured tier.
+            </p>
 
-        <div class="my-16 max-w-xl not-prose">
-          <Sponsors includeFeatured={true} />
-          <p class="text-sm text-gray-600 dark:text-gray-400 mt-4">
-            Join these amazing sponsors who make DDEV possible. <a href="/foundation" class="underline">Learn more about the DDEV Foundation</a>, our 501(c)(3) nonprofit organization.
-          </p>
+            <p>Each tier includes all benefits of lower tiers.</p>
```
stasadev: From my perspective, when I open the page, I simply don't read this sentence.

I would add this to each `rewards` (except the first one): "All the benefits of lower tiers".

stasadev:
BTW, GitHub supports this:

https://github.com/sponsors/ddev?amount=25

So links can be customized to preselect the amount.

ultimike:
Looking really good - some additional nits:

1. There seems to be a bit of a gap above "Sponsorship levels", no?
2. On the GitHub and Paypal buttons, can we make them a bit more consistent? Maybe change the text on both to "Sponsor via GitHub" and "Sponsor via PayPal". Also, can we add the same "credit card" icon thing to the GitHub button?
3. I'd get rid of the "Most popular" tag on the $100/month box. That's just silly BS marketing.
4. coder.ddev.com, diffy, as well as ddev.com and github should be linked to their respective URLs.
5. I'm tempted to suggest that we should add a "Direct billing" button to each box. I recognize that there's info about it lower on the page, and I'm not sure where we might link this new button to. Thoughts?

thanks,
-mike

rfay:
I just want to say @bmartinez287 - Despite the fact that everybody has opinions, you have taken the step to move this forward and that's key! Please don't be frustrated with the opinions. I think we'll get enough of them consolidated.

bmartinez287:
> I just want to say @bmartinez287 - Despite the fact that everybody has opinions, you have taken the step to move this forward and that's key! Please don't be frustrated with the opinions. I think we'll get enough of them consolidated.

No worries, quick feedback is the best kind. Specially since it gets it moving. It's a bit harder if is sits stale for too long.

bmartinez287:
the updated version has all suggestions except for a solution for direct billing.

ultimike:
@bmartinez287 Looks fantastic! I echo what @rfay wrote above about us all being really thankful for your work on this - I'll buy you a fancy drink 🍸 in Asheville in a couple of weeks!

Any chance you can drop the "Credit card" tag to the second line on the GitHub button (to match the PayPal button)?

Any thoughts on what we're going to do with all the other (non-$$$) info on the current https://ddev.com/support-ddev/ page?

-mike

stasadev:
Awesome!

One thought about the "Credit Card" label: do we actually need to show it? It feels like paying by card is the default expectation these days.

If we keep it on a second line, the button gets quite large. If we keep it on the same line, it starts to feel a bit text-heavy.

Also, if we end up adding a third button for direct billing, having some extra space available could be helpful.

ultimike:
I'm fine either way - my focus is consistency. 

-mike

bmartinez287:
Its updated. Not sure how to handle the direct billing and since that might not happen all the time maybe that can be added later?
Any other thoughts?

ultimike:
I think https://20260622-bmartinez-ddev-supp.ddev-com-front-end.pages.dev/support-ddev/ is good-to-go.

My only concern is that if this replaces https://ddev.com/support-ddev/, what happens to the rest of the non-$$$ content that is on the current page?

-mike

src/pages/sponsor.astro
```diff
@@ -1,130 +1,266 @@
...
+      "All the benefits of lower tiers",
+      'Your logo and link on [ddev.com/support-ddev](https://ddev.com/support-ddev/) and [github.com/sponsors/ddev](https://github.com/sponsors/ddev) in the "Featured Sponsors" section',
+      "Prioritized bug reports",
+      "DDEV Tip of the Day names and thanks you — seen by thousands of developers every day in the terminal",
+      "Sponsor spotlight blog post on ddev.com",
+    ],
+  },
+  {
+    amount: "1,000",
+    name: "Strategic Sponsor",
```
rfay, line 77: I love the name "strategic sponsor"

src/pages/support-ddev.astro
```diff
@@ -1,130 +1,266 @@
...
+      'Your logo and link on [ddev.com/support-ddev](https://ddev.com/support-ddev/) and [github.com/sponsors/ddev](https://github.com/sponsors/ddev) in the "Featured Sponsors" section',
+      "Prioritized bug reports",
+      "DDEV Tip of the Day names and thanks you — seen by thousands of developers every day in the terminal",
+      "Sponsor spotlight blog post on ddev.com",
+    ],
+  },
+  {
+    amount: "1,000",
+    name: "Strategic Sponsor",
+    description: "Great for large organizations.",
```
rfay:
```suggestion
    description: "Great for larger organizations.",
```

src/pages/support-ddev.astro
```diff
@@ -1,130 +1,266 @@
...
+
+const sponsorshipTiers = [
+  {
+    amount: "25",
+    name: "Freelancer Sponsor",
+    description: "Great for freelancers.",
+    rewards: [
+      "A heartfelt thank-you from the DDEV team",
+      "Sponsor badge on your GitHub profile",
+      "Universities and educational institutions: A one-time annual contribution of $400 qualifies for this tier. Multiple departments at the same institution can each sponsor independently.",
```
rfay: IMO this just adds unnecessary complexity. They can figure it out.
```suggestion
```

src/pages/support-ddev.astro
```diff
@@ -1,130 +1,266 @@
...
+      href: match[2],
+    }
+  })
+}
+
+const sponsorshipTiers = [
+  {
+    amount: "25",
+    name: "Freelancer Sponsor",
+    description: "Great for freelancers.",
```
rfay: We have oh-so-many lovely individual sponsors who just think it's the right thing to do even though they might not get their org to do it.

```suggestion
    description: "Great for freelancers and individuals.",
```

src/pages/support-ddev.astro
```diff
@@ -1,130 +1,266 @@
...
+    ],
+  },
+  {
+    amount: "100",
+    name: "Featured Sponsor",
+    description: "Great for small agencies.",
+    featured: true,
+    rewards: [
+      "All the benefits of lower tiers.",
+      "Your logo and link in the DDEV README",
```
rfay:
```suggestion
      "Your logo and link in the DDEV README and on ddev.com",
```

src/pages/support-ddev.astro
```diff
@@ -1,130 +1,266 @@
...
+      "Your entire team gets to use [Coder.ddev.com](https://coder.ddev.com/) to experience DDEV in the cloud. See announcement.",
```
rfay:
```suggestion
      "Your entire team gets to use [Coder.ddev.com](https://coder.ddev.com/) to experience DDEV in the cloud. See [description](/blog/coder-ddev-com-announcement/).",
```

rfay (changes requested):
This is really fantastic and we're almost there. I made some minor suggestions you can accept or edit on web. 

I would like to figure out a way to make the $500 and $1000 tiers more obvious that they should be done with invoicing. At this level if they do it via GitHub it's relatively expensive (they spend more than needed) and it doesn't work as well for their accounting system.  Consider swapping the GitHub button for an "Invoicing" button?

bmartinez287:
All the incoming changes were merged. Anything else?

rfay:
Thanks, do you see any way to emphasize invoiced support on the two higher options?

src/pages/contact.astro
```diff
@@ -80,6 +80,25 @@ const title = `Contact Us`
...
+            <li>Monthly amount or sponsorship tier</li>
+            <li>Preferred payment method (ACH, Wise, bank transfer, IBAN, SWIFT, or check)</li>
+            <li>Any purchase order, VAT ID, or tax forms (for example a W-9) you need</li>
+          </ul>
+          <p>
+            We'll reply with an invoice and payment details. As a US 501(c)(3)
+            nonprofit, sponsorships to the DDEV Foundation may be tax-deductible.
+          </p>
+        </section>
+
```
stasadev, line 101: @rfay, I think you'll want to change the wording here.

stasadev (approved):
Looks good to me.

I fixed a couple of inconsistencies and moved the page to `/sponsor`.

src/pages/contact.astro
stasadev: > do you see any way to emphasize invoiced support on the two higher options?

This is what I added for:

image

rfay:
We need to redirect from /sponsor-ddev to /sponsor then if changing the route. For some reason I can't push into this remote 

But need to add this to the "generic redirects" in public/_redirects

```
diff --git a/public/_redirects b/public/_redirects
index b073a9c..618c474 100644
--- a/public/_redirects
+++ b/public/_redirects
@@ -16,6 +16,8 @@

 # Generic redirects
 /board /foundation#board 301
+/sponsor-ddev /sponsor 301
+/sponsor-ddev/ /sponsor 301
```

stasadev:
I already added a redirect from `/support-ddev` to `/sponsor`.

There's no`/sponsor-ddev`.

rfay:
Well, let's get on with it! Congrats @bmartinez287 and all! Followups are welcome. Thanks to everybody for working this through and making it happen.

PR #666, opened 2026-07-10
blog: Add Claude Remote Control with coder.ddev.com

rfay:
## Summary

- New blog post walking through pairing Claude Code's Remote Control feature with a disposable coder.ddev.com workspace: disabling workspace autostop, running Claude in a named tmux session via `cc-remote-control.sh`, basic tmux usage, and following along from a phone.
- Small AGENTS.md addition: always write commit messages to a file and use `git commit -F <file>` instead of `-m "$(cat <<'EOF' ...)"`, which is unreliable in bash.

Rendered at https://pr-666.ddev-com-fork-previews.pages.dev/blog/claude-remote-control-with-coder/


rfay:
I think I'll do this over with an updated template that explicitly supports claude remote control

PR #688, opened 2026-07-24
chore: replace github.com/sponsors/ddev with ddev.com/sponsor

stasadev:
## The Issue

- For #655

Replaces `github.com/sponsors/ddev` with `ddev.com/sponsor`


rfay (approved):
Thanks!

PR #696, opened 2026-07-27
ci: add astro-link-validator for internal links, fixes #470

### LINKED ISSUE #470: Add link checker
We should have automated checks for all links similar to what we have in `ddev/ddev` repository.

1. MkDocs for checking internal links - find alternative for Astro
2. External links: use linkspector https://github.com/ddev/ddev/blob/main/.github/workflows/docscheck.yml

stasadev:
## The Issue

- Fixes #470

Adds `astro-link-validator` for internal links.

Adds more exclusions to linkspector.

### DIFF
--- a/.linkspector.yml
+++ b/.linkspector.yml
@@ -3,7 +3,7 @@ dirs:
 ignorePatterns:
   - pattern: "^mailto:"
   # Root-relative internal links (e.g. /contact, /img/...) aren't checked here;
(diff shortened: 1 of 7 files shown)


stasadev:
The fail is confirmed, which is good, now I can fix the internal links.

https://github.com/ddev/ddev.com/actions/runs/30284102671/job/90037478449?pr=696

rfay:
Thank you. Feel free to pull both when you like.

PR #704, opened 2026-07-29
layout: one width for every page, blog post sidebar, rebuilt footer, fixes #161, fixes #439

stasadev:
## The Issue

- Fixes https://github.com/ddev/ddev.com/issues/161: the hero and nav stayed at `max-w-4xl` while the viewport kept growing, so from 1024px up the extra space turned into void margin instead of page.

- Fixes https://github.com/ddev/ddev.com/issues/439: the homepage terminal was absolutely positioned at `right: -40rem` inside a `h-72` box, so on a 13" screen it hung off the side at the wrong size.

Both of those live in the shell of the page, so this takes on the inconsistencies around them at the same time. Every page picked its own width (`max-w-4xl`, `-5xl`, `-6xl`, `-7xl`) and its own gutter, so nothing lined up between blocks or between pages. The footer had grown into one flat list that mixed social accounts with pages and named several of them twice. A blog post opened with a full-width feature image, which meant scrolling before you could read the first paragraph, and the space beside the text was doing nothing.

## How This PR Solves The Issue

Preview base: https://20260729-stasadev-revamp.ddev-com-front-end.pages.dev/

| Page | Before | After | Preview | Live |
| --- | --- | --- | --- | --- |
| Every page | `max-w-4xl` (56rem) on most pages, `max-w-5xl`/`-6xl`/`-7xl` on the homepage's sections, and a different `px-*` gutter per block | `--site-width: 72rem` via `.site-container`, one `.site-gutter` (1.5rem, 2rem from 64rem up). Long-form text is capped separately at `--reading-measure` (42rem), wired into the typography plugin, so a wider page didn't make lines harder to read | [preview](https://20260729-stasadev-revamp.ddev-com-front-end.pages.dev/) | [live](https://ddev.com/) |
| Header | nav at `max-w-4xl`, logo sitting left of the content it heads | the site container and gutter, so the logo lines up with the page at every width (#161) | [preview](https://20260729-stasadev-revamp.ddev-com-front-end.pages.dev/) | [live](https://ddev.com/) |
[...]

PR #706, opened 2026-07-30
layout: measure the header nav rather than collapsing it at a breakpoint

stasadev:
## The Issue

Follow-up to the review on #704. The header nav and the blog post title both switched to their narrow layouts at `lg` (1024px). Browser zoom shrinks the viewport in CSS pixels, so at 110-125% a 1200px window falls under that and
the nav collapsed wholesale under the menu button.

`w-2/3` on the nav's right-hand wrapper capped its contents at two thirds of the container, which is why the threshold had to be that high.

The menu behind that button was a `div` only the script could reveal, so with JavaScript off there was no way to reach the nav links below 1024px at all.

The byline's edit link was anchored to the full 72rem `site-container` instead of the text, landing about 26rem past the end of it on posts with no feature image.

- Nav links come from one array rendered into both the bar and the menu. A script measures them and drops links from the end until the rest fit, moving each one into the menu. The button shows only when something is hidden.
  Without the script the markup still falls back to `hidden lg:flex`.
- The menu is a `details`/`summary`, so it opens with JavaScript off and the script keeps no open/close code of its own. The separate close button is gone; the summary toggles.
- Nothing in the bar may shrink, or the measurements come back compressed and report room that isn't there. The menu's width stays reserved while it's hidden, so the result can't oscillate.
- `.title-grid` puts the blog title beside its feature image from 52rem, with the image column at `clamp(16rem, 36%, 25rem)` so it shrinks instead of stacking.
- The byline row is capped at `--reading-measure`.

Preview: https://pr-706.ddev-com-fork-previews.pages.dev/

1. Drag the window from 1400px to 320px: links should leave the bar one at a time and appear in the menu, with the logo never compressed.
2. At 1200px, zoom to 110%, 125% and 150%: as many links as fit at each.
3. Compare `/blog/release-v1-25-3/` against a post with no feature image, at 1400px, 900px and 700px.
4. With JavaScript off: all links above 1024px, and below it the button opens the menu.

## Release/Deployment Notes

None.

### DIFF
--- a/src/components/Header.astro
+++ b/src/components/Header.astro
@@ -3,54 +3,100 @@
  * Global site header.
  */
 import { GITHUB_URL, DOCUMENTATION_URL } from "../const"
-import {
-  MagnifyingGlassIcon,
-  Bars2Icon,
-  XMarkIcon,
-} from "@heroicons/react/24/solid/index.js"
+import { MagnifyingGlassIcon, Bars2Icon } from "@heroicons/react/24/solid/index.js"
 import SponsorsBanner from "./SponsorsBanner.astro"
 import ThemeToggle from "./ThemeToggle.astro"
+
+// Links leave the bar from the end as it runs out of room, so this is also
+// least-important-last.
+const navItems = [
+  { label: "Get Started", href: "/get-started" },
+  { label: "Download", href: "/download" },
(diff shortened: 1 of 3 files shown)


rfay (approved):
This looks fantastic on all the ones I spot-checked, thank you!

PR #724, opened 2026-08-21
content(blog): umbraco on ddev

millnut:
Blog post on running Umbraco with DDEV

Rendered at https://pr-724.ddev-com-fork-previews.pages.dev/blog/umbraco-on-ddev/


rfay:
Awesome, thanks! rebased.

rfay:
https://pr-724.ddev-com-fork-previews.pages.dev/blog/umbraco-on-ddev/

src/content/blog/umbraco-on-ddev.md
```diff
@@ -0,0 +1,188 @@
...
+# .ddev/config.yaml
+web_extra_daemons:
+  - name: umbraco
+    command: "dotnet watch run --non-interactive --no-launch-profile"
+    directory: /var/www/html/MyProject
+```
+
+That puts `dotnet watch` under the web container's supervisord, so supervisord restarts it on failure. Overriding the compose command instead would displace DDEV's own entrypoint, and a build error would leave you with a container that is up but serving nothing.
+
+### TLS stops at the router
```
rfay:
```suggestion
### TLS is terminated on the router
```

src/content/blog/umbraco-on-ddev.md
```diff
@@ -0,0 +1,188 @@
...
+author: Lee Mills
+featureImage:
+  src: /img/blog/2026/08/umbraco-on-ddev.png
+  alt: DDEV and Umbraco logos joined by a plus sign
+categories:
+  - Community
+  - TechNotes
+---
+
+DDEV's `generic` project type will run anything that brings its own web server, and people already use it for Node and other stacks. Umbraco is a less travelled case: .NET, Kestrel, and SQL Server. I wanted to know how much of the DDEV experience still holds up there. Clone a repository, run one command, have a working site and database a few minutes later.
```
rfay:
```suggestion
DDEV's `generic` project type will run anything that brings its own web server, and people already use it for Node and other stacks. Umbraco is a less-travelled case: .NET, Kestrel, and SQL Server. I wanted to know how much of the DDEV experience still holds up there. Clone a repository, run one command, have a working site and database a few minutes later.
```

src/content/blog/umbraco-on-ddev.md
```diff
@@ -0,0 +1,188 @@
...
+
+Most of it holds up. The result is [umbraco-clean-ddev](https://github.com/millnut/umbraco-clean-ddev), which runs Umbraco on .NET 10 with SQL Server and Adminer, and gets a new developer going with a single `ddev setup`. What took the time was a handful of places where DDEV expected something I had not worked out yet.
```
rfay: Why `ddev setup` ?

src/content/blog/umbraco-on-ddev.md
```diff
@@ -0,0 +1,188 @@
...
+
+## What I wanted out of it
+
+Four things, and nothing more ambitious than that. One command for someone joining the project. Nothing installed on my own machine: no SQL Server, no .NET SDK, none of the project's other dependencies. Projects that stay out of each other's way. A way to take a backup from Umbraco Cloud and debug against it locally.
```
rfay:
```suggestion
I wanted four things things out of this, and nothing more ambitious than that. One command for someone joining the project. Nothing installed on my own machine: no SQL Server, no .NET SDK, none of the project's other dependencies. Projects that stay out of each other's way. A way to take a backup from Umbraco Cloud and debug against it locally.
```

Consider making these bullets. I like the way it is expressed, but bullets might help people to track to the solution.

src/content/blog/umbraco-on-ddev.md
```diff
@@ -0,0 +1,188 @@
...
+# .ddev/config.yaml
+name: umbraco-clean
+type: generic
+docroot: ""
+webserver_type: generic
+omit_containers: [db]
+disable_settings_management: true
+```
+
+`omit_containers: [db]` drops DDEV's MariaDB, since SQL Server runs as a separate service and MariaDB would only sit there idling. `disable_settings_management: true` stops DDEV writing a `php.ini` and an `nginx.conf` that nothing in the project would ever read.
```
rfay:
```suggestion
`omit_containers: [db]` drops DDEV's MariaDB, since SQL Server runs as a separate service and MariaDB would only sit there idling. 
```

`disable_settings_management` is only for explicit project types (like `typo3` etc), where DDEV helps with setting up CMS-specific settings for them.

src/content/blog/umbraco-on-ddev.md
```diff
@@ -0,0 +1,188 @@
...
+```yaml
+# .ddev/config.yaml
+web_extra_exposed_ports:
+  - name: umbraco
+    container_port: 80
+    http_port: 80
+    https_port: 443
+```
+
+Those two port numbers being equal matters. DDEV builds the Traefik backend URL from `http_port` rather than `container_port`, so an otherwise reasonable pairing of `container_port: 8080` with `http_port: 80` sends traffic to port 80 in the container and returns a 502. I lost a while to that one before reading the router configuration it generates.
```
rfay, line 60: I'd be interested in knowing more about what you experienced here. The router definitely knows how to use port 443/80 to send to a different http port (if your webserver were running on port 8080 for example)

src/content/blog/umbraco-on-ddev.md
```diff
@@ -0,0 +1,188 @@
...
+
+### The .NET SDK is not in the web image
+
+DDEV's web image has no .NET SDK, so `.ddev/web-build/Dockerfile` adds one. The important thing about that file is what it does not contain:
```
rfay: Recommend: 
```suggestion
DDEV's web image has no .NET SDK, so `.ddev/web-build/Dockerfile.dotnet` adds one. The important thing about that file is what it does not contain:
```

src/content/blog/umbraco-on-ddev.md
```diff
@@ -0,0 +1,188 @@
...
+
+```dockerfile
+# .ddev/web-build/Dockerfile
```
rfay:
```suggestion
# .ddev/web-build/Dockerfile.dotnet
```

src/content/blog/umbraco-on-ddev.md
```diff
@@ -0,0 +1,188 @@
...
+
+I also left `webimage_extra_packages` out of `config.yaml` on purpose. DDEV injects it ahead of the `web-build` Dockerfile, so any package listed there is looked up before the Microsoft feed exists.
+
+### SQL Server on Apple Silicon
+
+Full SQL Server publishes no ARM64 image, so the database service is Azure SQL Edge. It speaks the same TDS protocol and Umbraco cannot tell the difference.
+
+DDEV does have an add-on for SQL Server, [ddev-sqlsrv](https://github.com/ddev/ddev-sqlsrv), which runs full SQL Server and works on Apple Silicon through Rosetta 2 emulation. Apple has announced Rosetta is going away, and I wanted to see what was possible without installing it.
+
+This is the choice I would most like to revisit. If Microsoft ever publishes an ARM64 image, I would rather be running full SQL Server than Azure SQL Edge.
```
rfay, line 92: Could this be used in ddev/ddev-sqlsrv, or an equivalent add-on (one that's less ambitious than yours? I didn't know that this was even a possibility. It's a long-term surprise that MS continues to fail at providing ARM64 on so many things.

Maybe you can say a bit more about why azure-sql-edge can be used as a stand-in here.

src/content/blog/umbraco-on-ddev.md
```diff
@@ -0,0 +1,188 @@
...
+    healthcheck:
+      test:
+        - CMD-SHELL
+        - python3 -c "import socket; s=socket.create_connection(('localhost',1433),2); s.close()"
+      start_period: 30s
+```
+
+The healthcheck is a raw socket connection because Azure SQL Edge ships no `sqlcmd` to query with. It only proves the port is open, and the server accepts logins a moment after that, so the check is weaker than I would like. The 30 second grace period covers the initialisation of the system databases on a first run, which takes about 20 seconds and would otherwise exhaust the retries.
+
+Only `sqlserver` and `adminer` are defined in that compose file. The web container is configured entirely through `config.yaml`, so DDEV keeps ownership of its lifecycle.
```
rfay: > are defined in that compose file

I assume you men "in the azure-sql-edge image"?

src/content/blog/umbraco-on-ddev.md
rfay: Maybe you can link to these components as well. I hadn't heard of "Kestrel" before.

rfay (changes requested):
Very cool, testing it now. Minor comments and suggestions.

src/content/blog/umbraco-on-ddev.md
millnut: It currently copies the `appsettings.Local.json.example` file to `appsettings.Local.json` and runs `ddev add-on get ddev/ddev-adminer`. Mainly as a quickstart command so the local app settings has the unattended install config.

rfay: Thanks. It might be easier to be explicit about the `ddev add-on get` by putting it in a pre-start host hook (if it didn't exist).  Perhaps all the setup could be done like that.

src/content/blog/umbraco-on-ddev.md
```diff
@@ -0,0 +1,188 @@
...
+## Getting it running
+
+```bash
+ddev setup
+```
+
+That copies `appsettings.Local.json.example` into place, installs the Adminer add-on and starts the environment. On first boot Umbraco runs its unattended install, creating the schema and the default admin user, `admin@example.com` with password `1234567890`.
+
+:::warning[Error 4060 on first boot]
+The first boot logs `Cannot open database "UmbracoDb"` with error number 4060. The database does not exist until the unattended install creates it, so Umbraco logs the error once during startup. It does not recur, and you do not need to do anything.
```
rfay: Saw this for a while, "It does not recur" is not 100% correct, but fine here.

rfay (changes requested):
This is great, and I was able to get it going quite easily. Impressive usage. 
* I don't think the .ddev/.env relationship to the `appsettings.Local.json.example` are quite correct. Perhaps the created appsettings.Local.json doesn't respect the .ddev/.env in the future. 
* `ddev logs` and `ddev logs -f` are pretty useful and fun with this, it would probably be good to mention them.
* You may want to consider `performance_mode: none` in the config; it's my bet that mutagen is not useful for dotnet.

So much fun, thanks! When you think it's ready we'll publish.

src/content/blog/umbraco-on-ddev.md
millnut: Have replaced this would the pre-start as I wasn't aware you could have host level hooks. Adminer I've committed to save the `ddev add-on get` step as I already had to commit the compose file, so adding the command was a quick win

src/content/blog/umbraco-on-ddev.md
millnut: azure-sql-edge was retired by Microsoft, so is purely used as an experiment on how far I could get without enabling rosetta on macOS with Apple Silicon.

Given Microsoft's recent partnerships in the ARM space I'm hoping we'll see a true full SQL Server docker image with ARM64 support.

src/content/blog/umbraco-on-ddev.md
millnut: These are the compose service names, I can mention that `sqlserver` uses `azure-sql-edge` if that would make it clearer?

src/content/blog/umbraco-on-ddev.md
millnut: Yep have updated this part to reflect it occurs a few times until the app can connect

millnut:
Thanks for the feedback @rfay I've updated the repo with your suggestions and updated the blog content for re-review

src/content/blog/umbraco-on-ddev.md
millnut: This may be me not being familiar enough with DDEV's routing but this was the only way I could get it to serve Umbraco

src/content/blog/umbraco-on-ddev.md
rfay: I've been following the issue about MS ARM64 on sqlsrv for several years now. They don't ever chime in really. 
* https://github.com/microsoft/mssql-docker/issues/802

src/content/blog/umbraco-on-ddev.md
rfay: There really isn't anything that you should have to know about DDEV's routing. But sometimes an *application* is picky whether it can be found on the same container port as the router port (reverse proxy).

src/content/blog/umbraco-on-ddev.md
```diff
@@ -0,0 +1,223 @@
+---
+title: "Umbraco on DDEV: .NET, SQL Server, and the generic project type"
+pubDate: 2026-08-20
```
rfay:
```suggestion
pubDate: 2026-08-24
```

rfay:
Thanks! This is such a great tutorial example!

src/content/blog/umbraco-on-ddev.md
millnut: Ah yes that thread I've been watching that for awhile as well

millnut:
> Thanks! This is such a great tutorial example!

Thanks @rfay hopefully I can do a follow up when things stablise more on the SQL Server front

PR #733, opened 2026-09-02
content(usage-stats): note Amplitude event cap, move sponsorship section up

rfay:
## The Issue

- Related to ddev/ddev#8614

Amplitude changed its terms for free projects, so DDEV's usage tracking now exceeds the free event quota partway through most months, making the Amplitude-based stats on the usage-stats page incomplete with no indication to readers.

Adds a warning above the Amplitude-based stats linking to ddev/ddev#8614, and moves the DDEV Sponsorship section (unaffected, separate data feed) up to right after that notice instead of at the bottom.

Visit https://20260902-amplitude.ddev-com-front-end.pages.dev/usage-stats and confirm the warning and Sponsorship section now appear above the Amplitude charts.

### DIFF
--- a/src/pages/usage-stats.astro
+++ b/src/pages/usage-stats.astro
@@ -141,6 +141,79 @@ const sponsorshipGoal = sponsorshipData?.current_goal
           )
         }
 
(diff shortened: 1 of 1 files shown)


stasadev (approved):
Looks good to me.

